Issue #5, opened 2019-08-06
Should the manifest set minimum property requirements?


mattgarrish:
The web publications implementation of the manifest defined the required and recommended sets of properties. That didn't come across with part 1.

Should we add a section that sets a similar common base of properties for all manifests regardless of implementation?

If so, should we stop recommending properties and only define what is critical and leave it to implementations to recommend from the rest plus add whatever they need?

mattgarrish:
We do define something like I'm suggesting through the WebIDL, but you have to know to look at it. Better to have it clearly written out.

PR #18, opened 2019-08-07
Add requirements section to manifest definition

mattgarrish:
This PR is a start on addressing #5. It lists the two required properties (title and reading order) and indicates that all others are optional, but priority is also dictated by specific implementations.

Interestingly, the following is the minimum valid manifest, as the "required" properties are both compiled when not set:

```
{
    "@context" : ["https://schema.org", "https://www.w3.org/ns/wp-context"],
    "type"     : "CreativeWork"
}
```

I'm thinking it might be worth listing context and type as required properties, even though we didn't before, rather than assume them as part of a default configuration.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/18.html" title="Last updated on Aug 7, 2019, 8:01 PM UTC (4be22d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/18/c1c1d87...4be22d2.html" title="Last updated on Aug 7, 2019, 8:01 PM UTC (4be22d2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -192,6 +192,30 @@
 
 			</section>
 
+			<section id="manifest-requirements">
+				<h5>Requirements</h5>
+
+				<p>The following properties are REQUIRED in all implementations of the a <a>digital publication</a>
+					manifest:</p>
+
+				<ul>
+					<li>
+						<a href="#default-reading-order">default reading order</a>
+					</li>
+					<li>
+						<a href="#pub-title">title</a>
+					</li>
+				</ul>
+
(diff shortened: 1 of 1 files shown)


mattgarrish:
> Actually, the text requires contex and type, so I would agree adding them as required properties.

Right, we introduce them in a pedagogical way, sort of like walking through the authoring.

I've listed them in the requirements section and take the MUSTs out of their respective sections so that we don't duplicate conformance but have it all in one place.

PR #26, opened 2019-08-09
Clarify manifest discovery

mattgarrish:
Opening this PR as I noticed the manifest still seems closely tied to the PEP and linking to a manifest. To make things more (obviously) flexible for implementations of the manifest, it includes the following changes:

- renames "manifest association" to "manifest discovery" and adds section that other discovery methods than linking/embedding are allowed;
- adds to base url calculation that if another method is used, the format has to define the rules
- clarifies that html document input to lifecycle steps may not always be set

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/26.html" title="Last updated on Aug 9, 2019, 4:54 PM UTC (81c657f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/26/b733e29...81c657f.html" title="Last updated on Aug 9, 2019, 4:54 PM UTC (81c657f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -583,14 +583,14 @@ enum ProgressionDirection {
 						<p>The base URL for relative-URL strings is determined as follows:</p>
 
 						<ul>
-							<li>In the case of an <a href="#manifest-embed">embedded manifest</a>, the base URL is the
-									<a data-cite="!html#the-base-element">document base URL</a>&#160;[[!html]] of the
-								document that references the manifest;</li>
-							<li>In the case of a <a href="#manifest-link">linked manifest</a>, the base URL is URL of
-								the manifest resource.</li>
+							<li>In the case of an <a href="#manifest-embed">embedded manifest</a>, it is the <a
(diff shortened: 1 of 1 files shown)


mattgarrish:
I've added the links as part of formal references for what we want done in each case. Let me know if that's what you were thinking, or if you only wanted them as notes.

With respect to json-ld 1.1, does our use of "json-ld" mean we automatically roll over to the latest, or do we have to use json-ld11?

iherman:
That should be fine.

As for json-ld 1.0 vs. 1.1: not sure. At the moment, 1.1 has a different short name; this may change when the final Rec will come, and the final Rec may indeed overshadow 1.0. Actually, we have not really discussed this so far in the JSON-LD WG

Cc @BigBlueHat ?

mattgarrish:
I'm going to close this and leave an issue to return to the json-ld reference whenever that is settled.

PR #43, opened 2019-08-26
Adding isFamilyFriendly and Abridged

wareid:
Adding the two values mentioned in https://github.com/audiobooks/issues/12 

I am not sure if I needed to add something for the boolean, since these two are the first added to the document. Let me know and I'll correct that as well.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/43.html" title="Last updated on Sep 6, 2019, 3:50 PM UTC (87caf6b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/43/18242ea...87caf6b.html" title="Last updated on Sep 6, 2019, 3:50 PM UTC (87caf6b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -271,7 +271,7 @@ dictionary PublicationManifest {
              sequence&lt;LinkedResource>    resources = [];
              sequence&lt;LinkedResource>    links = [];
 };
-					
+
 
 dictionary CreatorInfo {
              sequence&lt;DOMString>         type;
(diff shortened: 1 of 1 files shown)


iherman:
> 2. how is "family friendliness" determined? Do we have to define it? (I'd be concerned if it's entirely subjective.)

I do not think we can/should define what it means. We may want to say that the judgement on what is family friendly greatly depends on cultures, jurisdiction, etc, and is therefore under the judgement of the publisher and/or local legislation.

iherman:
> Does it belong in the core specification, then? If we can't define how to use it, it becomes pretty random and unreliable. Nothing prevents the publisher from using other properties like this, or contentRating, if they find a purpose.

We just inherit, and surface explicitly, what schema.org has. Let that be their problem:-)

llemeurfr:
I think the question @mattgarrish rises is important and generic: what are the conditions for including a feature or metadata property inside the (core) WP spec, knowing that the manifest is extensible and may contain any property from any vocabulary? the only possible condition I see is "it is widely used by the industry and therefore requires standardization, for the sake of interoperability". 

The preview feature is certainly largely used and generic, partly implemented via a "preview" link already (for the full publication), and the addition of the associated "abridged" property (in the preview publication) is therefore logical. 

Is it the case for this "family friendly" property?  Are there "explicit" e-books?  Isn't it a very specific metadata inherited from the music industry and applied on audiobook "because it is required by the interchange format" ? What will happen to audiobook ingestion platforms if it is not set?

iherman:
> We also say in the extensibility section that any schema.org properties can be used, but it might make sense to make schema.org usage its own section just to surface it better.

👍 to that!

mattgarrish:
I'm going to merge this, as it's just the abridged property, and then work at the other issues I mentioned in https://github.com/w3c/pub-manifest/pull/43#pullrequestreview-279896543

Issue #39, opened 2019-08-21
Review of i18n self-review


iherman:
I have done the official i18n self-review in issue #38. Before adding the extra label to ping the i18n people, the WG should have a look at the individual items and, if possible, check whether I have made a mistake.

I have left two issues open:

- It should be possible to associate a metadata-type language declaration (which indicates the intended use of the resource rather than the language of a specific range of text) with multiple language values. [more](https://w3c.github.io/bp-i18n-specdev/#sec_lang_meta)
- Consider whether it would make sense to have one or more extra fields, in addition to the full name field, where users can provide part(s) of their name that you need to use for a specific purpose. [more](https://www.w3.org/International/questions/qa-personal-names#splitting)

We do not fulfill those, and the question is whether we have proper arguments to stay that way or whether we would add additional features accordingly. The translation, into our world, is:

- At the moment we have the `inLanguage`, defined as a _single_ language tag, and that signals the publication's language and the default language used for the metadata terms (title, etc). Do we need the possibility to have an _array_ of languages? What happens with multilingual ebooks, for example?
- At the moment the name of a person is one or several text terms (i.e., the person's name in English and Japanese). Do we need extra fields (an example is to specify the name used for sorting)? If so, how do we do that, knowing that having universal vocabularies for names can quickly become a very complicated nightmare...

I think, for both cases, the responses should be based on industry practice and business usage...

laudrain:
@iherman 
Multilingual books are for a monolingual audience. That's first example in [more](https://w3c.github.io/bp-i18n-specdev/#sec_lang_meta).
But it is useful to have transliterations in metadata, i.e. titles and/or author names of a Japanese book also transliterated in latin characters.

It is also a good practice to give in the metadata for the user library, author names in sorting order (Herman, Ivan ; Audrain, Luc), besides classical string (Ivan Herman; Luc Audrain).
That's a given in the ONIX feeds for digital webstores and libraries.

iherman:
> Multilingual books are for a monolingual audience. That's first example in more. But it is useful to have transliterations in metadata, i.e. titles and/or author names of a Japanese book also transliterated in latin characters.

... which, in fact, we do have, because the title term (`name` in the manifest) is an array of `LocalizableString`, i.e., one would add the title is as many languages/scripts as needed. So far so good.

[...]

PR #47, opened 2019-08-29
Fixes to language and direction of manifest/publication

mattgarrish:
This PR makes the following changes to address the issues raised in #39:

- allows for global manifest `language` declaration in the context
- adds a `direction` property for specifying global manifest text direction
- changes `inLanguage` to an array noting that first language is considered the primary for the publication
- merge `readingProgression` into `inDirection` and delete as it duplicates the same information
- updates the processing steps to check languages and directions conform to their specifications
- updates the webidl and schema

A couple of notes while making this change:

- There was a note referencing string-meta, but it wasn't clear if it was relevant to the manifest properties or to web publications. If you think it still belongs, @iherman, could you let me know where to put it back?
- I wasn't sure how strictly to define the context in the schema. It appeared to enforce order of context document URLs. I worked around that a bit by requiring unique entries, but there doesn't appear to be any way to allow random ordering and also enforce that 2/3 of the values be present.
- There's a bit of duplication in places that maybe we can find better workarounds for (e.g., the direction value descriptions repeat between manifest and publication, which I've used an include to reduce maintenance but still feels awkward).

Otherwise, let me know if there is anything else generally amiss with it.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 30, 2019, 5:20 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fpub-manifest%2Faca849f12f20f0a1f069a2b5345290f3ded22c94%2Findex.html%3FisPreview%3Dtrue)

```
Navigation Timeout Exceeded: 30000ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/pub-manifest%2347.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1378,7 +1378,7 @@ enum ProgressionDirection {
 								</tbody>
 							</table>
 
-							<p>The natural language MUST be a tag that conforms to&#160;[[!bcp47]], while the <dfn
+							<p>The natural language MUST be a well-formed in terms of &#160;[[!bcp47]], while the <dfn
 									data-lt="TextDirection">base language direction</dfn> MUST have one of the following
 								values:</p>
 
--- a/index.html
+++ b/index.html
@@ -151,7 +151,7 @@
 						<a>manifest</a> is defined by a set of properties that describe the basic information a user
 						agent requires to process and render the publication. For ease of understanding, these
 						properties are categorized as follows:</p>
-
+          
 					<dl>
 						<dt>
 							<a href="#descriptive-properties">descriptive properties</a>
@@ -172,7 +172,7 @@
 						</dd>
 					</dl>
 
-					<p class="note">These categorization have no relevance outside this specification (i.e., the
+					<p class="note">These categorizations have no relevance outside this specification (i.e., the
 						properties are not actually grouped together in the manifest).</p>
 				</section>
 
--- a/index.html
+++ b/index.html
@@ -151,7 +151,7 @@
 						<a>manifest</a> is defined by a set of properties that describe the basic information a user
 						agent requires to process and render the publication. For ease of understanding, these
 						properties are categorized as follows:</p>
-          
+
 					<dl>
 						<dt>
 							<a href="#descriptive-properties">descriptive properties</a>
@@ -359,7 +359,7 @@ enum ProgressionDirection {
 			<section id="manifest-lang-dir">
 				<h3>Manifest Language and Direction</h3>
 
-				<p>Each each natural language property value in a manifest (e.g., <a href="#pub-title">title</a>, <a
+				<p>Each natural language property value in a manifest (e.g., <a href="#pub-title">title</a>, <a
 						href="#creators">creators</a>) has a default natural <dfn>language</dfn>, which is the language
 					that it is expressed in (e.g., English, French, Chinese). It also has a natural <dfn>base
 						direction</dfn> in which it is written — the display direction, either left-to-right or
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -355,6 +356,158 @@ <h3>Manifest Contexts</h3>
...
+        "https://www.w3.org/ns/pub-context", 
+        {"language":"fr"}
+    ],
+    "name" : "Le roi René"
+}</code></pre>
+					</aside>
+
+					<p class="note">It is not necessary to specify the at-sign (<code>@</code>) in the key name when
+						declaring the language as the publication context maps the terms to each other.</p>
+
```
iherman: I am not sure this is something we should emphasize. While this is true, we do not do a similar remark v.a.v., e.g., `type` vs. `@type`. Maybe it is worth to emphasize, on the other hand, that order of these is important, ie, that the `language` should follow the pub context (which, technically, does these mappings).

index.html
```diff
@@ -355,6 +356,158 @@ <h3>Manifest Contexts</h3>
...
+					<p>The value of the <code>language</code> MUST be set to a language code as defined in
+						[[!bcp47]].</p>
+
```
iherman: I think the i18n people prefer the formalism: "to a language code that is well-formed as defined in [[!bcp47]]"

index.html
```diff
@@ -355,6 +356,158 @@ <h3>Manifest Contexts</h3>
...
+        "name" : {
+            "value"    : "Marcel Proust",
+            "language" : "fr"
+        }
+    }
+}
+</pre>
+					<p>The value of the <code>language</code> MUST be set to a language code as defined in
+						[[!bcp47]].</p>
+
```
iherman: Same remark on the "well-formed" for the language tag.

index.html
```diff
@@ -355,6 +356,158 @@ <h3>Manifest Contexts</h3>
...
+        "https://schema.org",
+        "https://www.w3.org/ns/pub-context",
+        {"language": "fr"}
+    ],
+    "type"       : "Book",
+    &#8230;
+	"author"     : "Marcel Proust",
+    &#8230;	
+}
+</pre>
```
iherman: Something went wrong with the formatting of the "author" line, its indentation is too large (probably the tab character must be exchanged against spaces)

index.html
```diff
@@ -355,6 +356,158 @@ <h3>Manifest Contexts</h3>
...
+					<p>See <a href="#value-objects"></a> for more information.</p>
+
+					<p>It is <em>not</em> possible to set the direction explicitly for a specific value.</p>
+
```
iherman: I would think (reflecting to one of your questions in the preamble) that this is a good place for the reference to string-meta

index.html
```diff
@@ -1236,34 +1389,66 @@ <h5>Duration</h5>
...
+							language direction MUST have one of the following values:</p>
 
-						<p>The natural language and base direction can be set for both the <a
-								href="#manifest-default-language-and-dir">publication</a> and the <a
-								href="#manifest-specific-language-and-dir">natural language properties values</a> of the
-							manifest.</p>
+						<ul data-include="common/html/dir-values.html" data-include-replace="true"></ul>
 
 						<p>If a user agent requires the language and one is not available in the manifest (either
 							globally or specifically for that property), or the obtained value is invalid, the user
```
iherman: Probably worth emphasizing 'publication language', rather than just 'language'

Also, when referring to "the non-empty language" it is better to refer explicitly to 2.5.1

index.html
```diff
@@ -2843,6 +2887,30 @@ <h2>Processing a Manifest</h2>
 								object in which one or more properties expect entities.</p>
 						</li>
 
+						<li>
+							<p>(<a href="#manifest-lang-dir-global"></a> and <a href="#manifest-lang-dir-local"></a>)
+								Recursively check that every <var>language</var> property in <var>processed</var>
+								conforms to [[!bcp47]] and issue a warning for any invalid values found.</p>
+						</li>
```
iherman: conforms -> is well-formed

schema/publication.schema.json
```diff
@@ -6,7 +6,26 @@
...
+            "items": [
+	            {
+	            	"type": "string",
+	            	"enum": ["https://schema.org", "https://www.w3.org/ns/pub-context"]
+	            },
+	            {
+	            	"type": "string",
+	            	"enum": ["https://schema.org", "https://www.w3.org/ns/pub-context"]
+	            },
+	            {
```
iherman: Actually, the pub-context should follow the schema.org one, because it adds some extra features to schema.org properties... better keep to the order

iherman:
@mattgarrish this looks, in general, really good, all my comments are superficial...

As for your questions:

- I have put a comment for string-meta. That comment is definitely valid for the manifest, and not for the comment files
- As I said, it is better to impose a strict order. The language one must follow the pub context, that is for sure (otherwise 'language' may not be understood); for the other two I am not 100% sure, and better safe than sorry...
- Yeah, the direction value is a duplication but...

That being said, I wonder whether the 'auto' value makes any sense for `inDirection`. The auto value may be useful if looking at an individual text item, but the `inDirection`'s usage is much more abstract, because it is meant to, say, help the user interface design (placing of menus, that sort of things). I would think that we should remove 'auto' there, just as the reading progression direction does not use 'auto' either.

(Which may mean that the term `ProgressionDirection` in the WebIDL may be a misnomer...)

iherman:
(Any idea why there are merge/rebase conflicts?)

mattgarrish:
> (Any idea why there are merge/rebase conflicts?)

There was a conflict because of the merge of the schema.org section from the other PR, but I resolved that last night. I'm not seeing the issue anymore.

schema/publication.schema.json
mattgarrish: Okay, then we should probably make this a requirement in 2.5.

mattgarrish:
> Which may mean that the term `ProgressionDirection` in the WebIDL may be a misnomer...

Hm, isn't this just epub's page progression direction, which we recently decided we weren't going to add?

Are there languages in which you read the text ltr but move rtl from page to page (or vice versa)? If not, then doesn't inDirection also serves as an indicator of page progression?

iherman:
>> Which may mean that the term ProgressionDirection in the WebIDL may be a misnomer...
>
> Hm, isn't this just epub's page progression direction, which we recently decided we weren't going to add?
>

To be honest, I do not know, I did not pay attention to that EPUB discussion. If it was rejected there, then you should propose removing here, too!

> Are there languages in which you read the text ltr but move rtl from page to page (or vice versa)? If not, then doesn't inDirection also serves as an indicator of page progression?

You are probably right. Before this PR `inDirection` had a dual role, which justified having something separate, but that may be gone now.

iherman:
> > (Any idea why there are merge/rebase conflicts?)
> 
> There was a conflict because of the merge of the schema.org section from the other PR, but I resolved that last night. I'm not seeing the issue anymore.

Actually, I still see the conflict on my screen for the PR... :-(

index.html
```diff
@@ -355,6 +356,167 @@ <h3>Manifest Contexts</h3>
 						<code>https</code> when referencing Schema.org in the manifest.</p>
 			</section>
 
+			<section id="manifest-lang-dir">
+				<h3>Manifest Language and Direction</h3>
+
+				<p>Each each natural language property value in a manifest (e.g., <a href="#pub-title">title</a>, <a
```
llemeurfr: Each each = typo

mattgarrish:
> If it was rejected there, then you should propose removing here, too!

That's the weird thing. It starts off mentioning progression across resources, but then reading the definition it's actually almost a verbatim copy of `inDirection`. I wonder if we lost track that we had directionality defined there, too, because of the name.

At any rate, I'm going to push a change that merges what useful bits there seemed to be in `readingProgression` into `inDirection`, including keeping the `ltr` default.

llemeurfr:
IMO the `readingProgression` -  which drives RS behaviour related to the use of left-right arrows or user taps on the left of right of the screen - should stay away from the direction of textual information in the manifest. 

Two clues: 
- readingProgression will have new values later, when digital comics are integrated: top-to-bottom is obvious (webtoons) and bottom-to-top will be useful also (this is not the place to discuss this).  
- in an audiobook the is no readingProgression but there is still a direction for textual information in the manifest.

mattgarrish:
> Actually, I still see the conflict on my screen for the PR... :-(

No clue. Must be a github bug. I have the big green merge button on my end, so I should be good update.

I'd like to get this merged, too, in the interests of getting the draft as complete as we can for the next call. Any objections to moving ahead with that by end of today?

llemeurfr:
I commented to quickly; I see now the proposal for a separate `direction` (for textual metadata) and ' inDirection' (for page progression). I would prefer keeping `readingProgression` and suppressing `inDirection` then, as the latter is confusing when compared with `direction`.

iherman:
@llemeurfr yep, we wanted to remove duplication.

However... semantically, `inDirection` is the pair to `inLanguage`, ie, giving information about the language and base direction to be expected for the publication resources. I have the impression that what you say about, say, comics, is different, which for me is an argument to keep it after all rather than try to push the semantics on `inDirection`.

Editorially, I guess your example is very good. One may have `progressionDirection` being top to bottom, but the `inDirection` set to 'ltr' the individual graphics in the comics are set to left-to-right.

Ideally, `progressionDirection` should get more values right now, to make the distinction clearer...

mattgarrish:
If we add top to bottom, which matches with css writing modes, then indirection would be a bad choice.

So we discard indirection, correct, as what value does the text flow within documents matter so long as a progression is known? It's more a concern for manifest values.

iherman:
> If we add top to bottom, which matches with css writing modes, then indirection would be a bad choice.

I think my idea was 

- `direction` should stay as it is, it is related to the base text direction
- `readingProgression` should remain, but with possibly a richer set of values as now

@llemeurfr is this what you had in mind?

iherman:
Actually, just to move ahead, I would propose:

- keep `direction` as is with a textual direction, and `readingProgression` intact
- *merge this Pull Request*
- let us have a separate issue on whether we want a richer set of values for `readingProgression`

Otherwise we are mixing issues, which will lead to a mess

mattgarrish:
Right, I wasn't suggesting we remove `direction` since it's actually used for presentation. I'm questioning what value `inDirection` has. It's not in EPUB, for example, and hasn't been raised as a concern there before. But for another issue.

PR #53, opened 2019-09-02
Add security and privacy considerations

mattgarrish:
May not stand up to review, but here's some starter prose.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/53.html" title="Last updated on Sep 4, 2019, 3:50 PM UTC (b0c2959)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/53/356247f...b0c2959.html" title="Last updated on Sep 4, 2019, 3:50 PM UTC (b0c2959)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2983,13 +2983,26 @@ enum ProgressionDirection {
 		<section id="security-privacy">
 			<h2>Security and Privacy Considerations</h2>
 
-			<p>Since this specification does not define an implementation of a publication manifest, security and
-				privacy considerations are generally left for each <a href="#extensions">profile</a> to identify.</p>
+			<p>Since this specification does not define an implementation of a publication manifest, specific security
+				and privacy considerations are left for each <a href="#extensions">profile</a> to identify.</p>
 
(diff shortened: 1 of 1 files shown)


iherman:
I think it is a good start to have this, but we may have to consider more text as for the specific terms that are used and proposed by this spec. Does the inclusion of personal data (name, URL-s, identifiers) come as a privacy risk? Although I do not think that, at the end of the day, it does, the fact that we do think about it is important.

One way of considering it to see if there are privacy related consideration in schema.org that we can simply refer to (after all, we do not have extra terms in this area.)

mattgarrish:
I've added a bullet list with some additional considerations. There are lots of possible security considerations for an actual format, but I can't think of what else to say at this level. Hopefully others will chime in.

I can't find anything about schema.org and security. As a pure vocabulary, it's probably not a concern how people expose their information.

iherman:
It now reads a bit funny... the first sentence says that we have nothing to say, and then we list some privacy and security considerations later in the same section...

mattgarrish:
Sure, ending off with a punt on specifics probably makes more sense now. I just pushed a minor reformulation to that end.

Issue #33, opened 2019-08-15
Add linked resource property called 'alternative'


marisademeglio:
`LinkedResource` needs a property called `alternative`, which has as its value a URL of a resource that represents the `LinkedResource`'s content in an alternate modality. It's used as follows:

1. audio book with mp3s in the reading order;  a reading order entry's `alternative` property points to a text file for that audio (see https://github.com/w3c/pwpub/issues/44)

2. audio book, as above, but with `alternative` pointing to a synchronized narration document

3. text book with HTML files in the reading order; `alternative` points to a synchronized narration document.

The resource pointed to by `alternative` appears in the `resources` list and is processed according to mime type. 

Reference example: draft of how to incorporate synchronized narration: https://w3c.github.io/sync-media-pub/packaging.html

llemeurfr:
Plus audio books with alternative bitrates.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-08-19-pwg#section8).

- `RESOLVED:  there is an in principle agreement with the proposal, a PR will follow with the details`
<details><summary><i class="differentiate">View the transcript</i></summary>
Add linked resource property called ‘alternative’<br/>
Luc Audrain: See Issue #33<br/>
Wendy Reid: this is to facilitate the use case for synch media<br/>
… we realized that there wasn’t anything in synched media that allowed for this so I suggested this issue be opened against manifest<br/>
… this affects more than just audiobook<br/>
Ivan Herman: this is the written version of an audio track?<br/>
… it’s value is the url to a resource?<br/>
… but it must be present in the links in the manifest?<br/>
… I imagine it could be done automatically in those transformations<br/>
Marisa DeMeglio: I like the idea of not having it be strict<br/>
Matt Garrish: There’s no reason it couldn’t be an array as well as a value, correct?<br/>
Wendy Reid: yeah, I think you could have multiple alternatives<br/>
Ivan Herman: I am backpedaling<br/>
… if I do it automatically then the media type will not be clear - especially if we do what Matt says<br/>
… I don’t think we have an automatic deduction of the media type<br/>
Laurent Le Meur: See Example of a similar structure in Readium<br/>
Laurent Le Meur: See Example for usage of a similar construction<br/>
Laurent Le Meur: there are some potential implementations on the market<br/>
… I support Ivan’s proposals<br/>
[...]

PR #56, opened 2019-09-09
Feature/add alternative

marisademeglio:
Adds a property to LinkedResource called `alternate`. 

Addresses https://github.com/w3c/pub-manifest/issues/33

***
<a href="https://pr-preview.s3.amazonaws.com/marisademeglio/pub-manifest/pull/56.html" title="Last updated on Sep 9, 2019, 9:43 PM UTC (3afedcd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/56/cd05f1a...marisademeglio:3afedcd.html" title="Last updated on Sep 9, 2019, 9:43 PM UTC (3afedcd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2787,10 +2787,10 @@ enum ProgressionDirection {
 }</pre>
 							</details>
 						</li>
-
+	
 						<li>
 							<p>(<a href="#value-link"></a>) If <var>term</var> is a <a
-									href="#resource-categorization-properties">resource categorization property</a> and
+									href="#resource-categorization-properties">resource categorization property</a> or an <code>alternate</code> property, and
 								any value in the array in <var>current</var> is a string, convert each string
 									<var>str</var> to an object with the following properties:</p>
 							<ul>
@@ -3169,7 +3169,17 @@ enum ProgressionDirection {
 							<td>
 								<p>References to one or more reformulation(s) of the resource in alternative formats.</p>
 								<p>When specified, <code>encodingFormat</code> indicates the format of the reformulation.</p>
+								<p>Order is not significant.</p>
 								<p>OPTIONAL</p>
+								<p>One or more of:</p>
+										<ul>
+											<li>a string, representing the <a>URL</a> of the resource reformulation in an alternative format; or</li>
+											<li>an instance of a <a href="#publication-link-def"
+														><code>LinkedResource</code></a> object</li>
+										</ul>
+										<p>The order of items is <em>not significant</em>. Non-HTML resources SHOULD be expressed as
+												<code>LinkedResource</code> objects with their
+												<code>encodingFormat</code> values set.</p>
 							</td>
 							<td>
 								<p>One or more instances of <a href="#publication-link-def"><code>LinkedResource</code></a> objects.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3173,13 +3163,19 @@ <h3><code>LinkedResource</code> Definition</h3>
...
+								<p>References to one or more reformulation(s) of the resource in alternative formats.</p>
+								<p>When specified, <code>encodingFormat</code> indicates the format of the reformulation.</p>
+								<p>OPTIONAL</p>
+							</td>
+							<td>
+								<p>One or more instances of <a href="#publication-link-def"><code>LinkedResource</code></a> objects.</p>
+							</td>
+							<td>
+								<a href="#value-array">Array</a> of <a href="#value-link">Links</a>
 							</td>
```
iherman: Is the order significant or not? This must be made clear in the spec, see, for example, the definition of reading order (2.7.3.2) if it is not the case.

index.html
```diff
@@ -3173,13 +3163,19 @@ <h3><code>LinkedResource</code> Definition</h3>
...
 							<td>
-								<a href="#value-array">Array</a> of <a href="#value-url">URLs</a>
+								<p>References to one or more reformulation(s) of the resource in alternative formats.</p>
+								<p>When specified, <code>encodingFormat</code> indicates the format of the reformulation.</p>
+								<p>OPTIONAL</p>
+							</td>
+							<td>
+								<p>One or more instances of <a href="#publication-link-def"><code>LinkedResource</code></a> objects.</p>
+							</td>
+							<td>
```
iherman: In other, similar terms (e.g., resource links) the definition allows the usage of single strings instead of a full-blown Linked Resource if the target is an HTML file. See the definition of resource list (see, e.g., 2.7.3.2). I believe that, for consistency's sake, the same should apply here.

If this is indeed so, then the step 5.d in the processing steps (section 5) should also be amended, because this property is not a 'resource categorization property', but the same processing should also apply.

index.html
```diff
@@ -3209,7 +3205,7 @@ <h3><code>LinkedResource</code> Definition</h3>
              sequence&lt;DOMString> rel;
              DOMString           integrity;
              double              length;
-             sequence&lt;DOMString> alternative;
+             sequence&lt;LinkedResource> alternate;
```
iherman: The tabulation of the WebIDL seems to be off...

index.html
```diff
@@ -3173,13 +3163,19 @@ <h3><code>LinkedResource</code> Definition</h3>
...
+								<p>OPTIONAL</p>
+							</td>
+							<td>
+								<p>One or more instances of <a href="#publication-link-def"><code>LinkedResource</code></a> objects.</p>
+							</td>
+							<td>
+								<a href="#value-array">Array</a> of <a href="#value-link">Links</a>
 							</td>
 							<td> (None) </td>
 						</tr>
```
iherman: I think an example for this property should be added to the spec, to highlight the role of alternate, mainly if a full array is used.

iherman:
Also, the JSON schema in https://github.com/w3c/pub-manifest/blob/master/schema/link.schema.json should also be changed (actually, regardless of whether this PR is accepted or not, because it is missing even in the old format!)

iherman:
Note to myself: the 'alternate' property should also be added to the vocabulary and the context file, because it is not a schema.org property after all.

index.html
marisademeglio: I think the order doesn't matter.

index.html
marisademeglio: Agree

index.html
marisademeglio: I added one

index.html
iherman: I am just knit-picking here:-)

In of 5.d there is a link to the resource categorization properties; I guess, for consistency, there should be a link to where the `alternate` is defined, too...

iherman:
> Also, the JSON schema in https://github.com/w3c/pub-manifest/blob/master/schema/link.schema.json should also be changed (actually, regardless of whether this PR is accepted or not, because it is missing even in the old format!)

@marisademeglio, if you cannot do it, I am happy to do it if and when this is merged. I cannot add it to the current PR, because it is in your cloned repository to which I do not have write access...

index.html
marisademeglio: Agree. In general, there aren't links to properties which are at the same level as `alternate`, so I didn't introduce this, but there are a few spots where I would think it would make sense.

marisademeglio:
The schema on my fork has `alternate` (🙏 @mattgarrish ): https://github.com/marisademeglio/pub-manifest/blob/feature/add-alternative/schema/link.schema.json#L60

I don't know what changes you have in mind though (you mentioned old format, I'm not so familiar with the history of this file) so if it needs more work, feel free to make the edits or let me know and I'll do it.

iherman:
Oops, sorry @marisademeglio, I missed that one in the schema file somehow. What you have is perfect, apologies for the noise!

index.html
marisademeglio: Well, one spot, anyway. The link has been added.

Issue #57, opened 2019-09-09
How Metadata Works in the Publishing World


dauwhe:
In much of the EPUB world, the metadata that matters is not inside the EPUB, but outside (in the form of ONIX). The metadata inside EPUBs is often wrong, is difficult to change, and there is very little incentive to make it accurate since it's mostly unused. 

In the web world, page metadata directly affects search ranking, Google rich snippets, etc. There is no out-of-band transmission of metadata. There is strong incentive to make it accurate.

How do we avoid the situation with EPUB, where we've spent decades worrying about metadata, continually changing how it's expressed, without really benefiting users?

mattgarrish:
This is generally why its a bad idea for publication specifications to dive so deeply into metadata vocabularies.

We wanted to provide a framework for metadata expressions with EPUB 3, but got sucked into the metadata vortex of despair by introducing some "essential" metadata that didn't seem to exist but that looked like the non-ONIX folks would need. That's led to EPUB being looked at as having to define the essential metadata, when metadata expressions really should be figured out at the publishing/publication level.

I opened https://github.com/w3c/wpub/issues/429 in part because I see a lot of the same happening here. The more we layer in the more it looks like what we exclude doesn't matter, and that leads to more requests to include things. Plus the more we recommend for certain areas of publishing the more annoying we make metadata for others.

The manifest is _somewhat_ unencumbered now that most metadata is optional, but we still define a whole lot of concepts that really aren't essential to user agents (dates, etc.).

During 3.1, we started to look at defining prescriptive metadata guidelines for publishers using alternative means, like best practices documents. Call me old fashioned, but it still strikes me as the best balance. Let each community define what it wants to express and how it wants to express it outside the specification.

iherman:
Well... we should be careful. The presence of, e.g., ONIX is clearly important for trade publishing. But we also know from our discussions that, at this moment, Publication Manifests will not be considered by trade publishing for some times which will keep to EPUB 3.x. 

What is the situation in other areas? E.g., the little I know about scholarly, where (at least for journals) the "publication" content is not dominated by packaging, which put things in a very different perspective.

mattgarrish:
> The presence of, e.g., ONIX is clearly important for trade publishing. But we also know from our discussions that, at this moment, Publication Manifests will not be considered by trade publishing for some times which will keep to EPUB 3.x.

Right, I'm not suggesting we pick a side in that debate.

[...]

Issue #16, opened 2019-02-14
Specification's version in the manifest


gregoriopellegrino:
Since we created a first propotype of Web Publication based on the current specifications, I wondered how to indicate in the manifest that it refers to the current specifications alpha or beta version.

In general, if in a few years we will define Web Publication 2.0, how will a User Agent distinguish current Web Publications from future ones?

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-09-09-pwg#section4-9).

- `RESOLVED:  close #16 without action (not adding version numbering)`
- `ACTION:  add an informational text on using dcterms (Matt Garrish)`
<details><summary><i class="differentiate">View the transcript</i></summary>
5.1. Issue #16 Specification’s version in the manifest<br/>
Garth Conboy: See Issue #16<br/>
Garth Conboy: One other wasn’t postponed, and hasn’t been addressed for some time - issue 16 - to do with version in the manifest.<br/>
… is this obviated or are we still live?<br/>
… we didn’t get to a satisfying conclusion…<br/>
<b>all:</b> silence<br/>
Wendy Reid: I do love the aspirational nature of the question<br/>
Charles LaPierre: Looking at this, because it also goes in line with the accessibility specification, with the conforms-to suggestion, I like that you can specify the version number and have all the information right there…<br/>
… we have that already as a basis for conformance to the a11y spec…<br/>
… but we’re trying to avoid compatibility issues, if we have a version in the future which requires special handling, if we don’t put it in there now we could be in trouble…<br/>
Dave Cramer: I hope we can avoid having version numbers on this.<br/>
Wendy Reid: +1<br/>
Dave Cramer: it feels like a nightmare of incompatibility and writing special paths and code etc.…<br/>
… we’ve managed to avoid it in HTML and CSS, let’s really try not to go there…<br/>
Garth Conboy: One can also argue that even if we do choose to go there, we don’t have to go there at the first try…<br/>
Ivan Herman: I agree - if we look back at the comments; if we want to express conformance with a version of the specification, there is a comment I made in Feb which said you can put this kind of information in the manifest…<br/>
… there’s a way of using dc terms to do that…<br/>
… if this information is necessary in the document, we can do that with the manifest…<br/>
Proposed resolution: close #16 without action (not adding version numbering) (Garth Conboy)<br/>
Wendy Reid: +1<br/>
Ivan Herman: +1<br/>
[...]

PR #59, opened 2019-09-12
Add basic dcterms:conformsTo example

mattgarrish:
One question I had is that there was already an example of `dc:subject` but the note before the example incorrectly suggested `dc` maps to dcterms.

Did we actually want to include a dc element set example, or is my correcting the prefix to `dcterms` fine?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/59.html" title="Last updated on Sep 15, 2019, 3:18 PM UTC (7790daf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/59/a514e50...7790daf.html" title="Last updated on Sep 15, 2019, 3:18 PM UTC (7790daf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2124,13 +2124,13 @@ enum ProgressionDirection {
 								href="https://www.w3.org/TR/json-ld/#compact-iris">Compact IRIs</a>&#160;[[!json-ld]],
 							with prefixes defined as part of the context.</p>
 
-						<p class="note"> A prefix definition <code>dc</code> for [[dcterms]] is included in the context
+						<p class="note">A prefix definition <code>dcterms</code> for [[dcterms]] is included in the context
 							file of [[schema.org]]. This means that it is not necessary to add the prefix explicitly.
 							The same is true for a number of other public vocabularies; see the <a
 								href="https://schema.org/docs/jsonldcontext.json">schema.org context file</a> for
 							further details.</p>
 
-						<pre class="example" title="Usage of the schema.org 'copyrightYear' and 'copyrightHolder' terms, as an extension to the basic data">
+						<pre class="example" title="Extending the basic data using the schema.org 'copyrightYear' and 'copyrightHolder' terms.">
 {
 "@context"        : ["https://schema.org","https://www.w3.org/ns/pub-context"],
 "type"            : "TechArticle",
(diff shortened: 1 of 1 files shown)


mattgarrish:
> Although, afaik, DCMI prefers users to use `dcterms:` I am not sure we have to make this decision in our spec.

Yes, but do we really need to get into this in the specification by introducing both when all we're aiming to do is show that other properties can be used. As in #57, if there's a need for the dc element set, it can be documented elsewhere.

EPUB banned the `dc:` prefix for properties since the restrictions on the elements wouldn't extend to the properties. Do the elements transform forward to dc or dcterms properties... we don't even know. Do they get treated equivalently, does one set take priority?

Given that dcterms exists for linked data, I'm perfectly fine having it as the only example prefix.

iherman:
To very honest, I don't have strong feeling about this. I see that there is a similar discussion on the epub repo; I am perfectly fine aligning ourselves on Epub.

llemeurfr:
Well, I can only say that reading `dcterms:subject` - even if technically ok - seems odd to me. I may be the only one. 

nb: In the current DCMI spec I don't clearly see a preference for properties from the dcterms namespace; I agree with Matt that the presence of both `dc:subject` and `dcterms:subject`in DCMI without assertion about their equivalence is odd.

mattgarrish:
Could we maybe avoid this issue entirely and use some other prefix example?

What about using a `bibo:` example and drop the `dc:subject` one?

I don't have a preference for `dcterms` over `dc` so much as I'm just hoping to avoid this whole two DC vocabulary issue.

tcole3:
As defined in schema.org's context doc, the dc:: prefix is tied to the (legacy) 2001 Dublin Core 1.1 elements namespace (http://purl.org/dc/elements/1.1/) while the dcterms: prefix is tied to the 'living' Dublin Core terms namespace (http://purl.org/dc/terms/). This is the convention generally followed by the library community. There have been no changes (no new elements, no changes in definitions) in the dc: namespace since 2001, while there have been many changes and updates to the dcterms: namespace over the last 2 decades.  

Also, as used in the context of the Open Archives (OAI-PMH) and other legacy XML applications using Dublin Core, the practice has been that properties in the dc: namespace cannot have any children (e.g., so can't break name into family and surname) and do not have any sub-properties or attributes, while more flexibility is assumed for properties in the dcterms: namespace.  

The latter are legacy conventions of course and not ones we (or schema) necessarily have to follow when serializing in json-ld, but I think it would occasion less confusion overall if we stuck exclusively in our examples to dcterms: and avoided any examples making use of dc:. The dc: 1.1 elements namespace is outdated and is generally seen in the library domain as not the best choice for Dublin Core in LD. 

And though I personally could live with @mattgarrish's suggestion to avoid dc:/dcterms: by using bibo: for our examples, among US libraries at least, I think many would find a bibo: example more confusing then a dcterms: example. 

Understanding that there will always be a few folks who prefer the fixity of dc:...

tcole3:
And technically, dcterms:subject is a refinement of dc:subject, see: https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-subject rather than being exactly equivalent...

mattgarrish:
> I think many would find a bibo: example more confusing then a dcterms: example.

@tcole3 I'm not suggesting we use `bibo:` for examples anywhere else in the specification, just to be clear. These specific examples are in the section on how to extend the manifest using prefixes from the schema.org context file. I don't know that we need two examples of dublin core there, especially, as you say, if `dcterms:` is the preferred vocabulary to use.

Let's keep the new example of `dcterms:conformsTo`, since there's some desire to show this for versioning, not just as a prefix example. But let's add a second example that says here's how you could use X to add some other property, where X can be any other prefix than `dc:` (it could be `foaf` if that's more palatable; I just threw `bibo` out as it sometimes comes up).

iherman:
Using dcterms for conforms to seems to be necessary, and is in the document due to #16. Seeing the discussion maybe the best is to keep to Matt's original proposal and use, in the examples dcterms only.

Maybe it is worth, in the note referring to schema.org, citing more of the namespaces that are predefined, like bibo, foaf, and indeed dc.

mattgarrish:
> Maybe it is worth, in the note referring to schema.org, citing more of the namespaces that are predefined

Works for me.

mattgarrish:
I've updated the note to add more of the vocabularies.

I also noticed that we don't actually show an example of extending the context but. I put in a fictitious example, but maybe someone can think of something better.  Let me know if I didn't get the prefix declaration right, too.

iherman:
Ship it:-)

Issue #65, opened 2019-09-16
Validate expected value types


mattgarrish:
While we validate some specific values for syntax (e.g., dates), we don’t say anything about what to do if a property doesn’t have its expected value, for example:

```
   “author”: true
```

would silently slip through the algorithm without warning, even though it is supposed to be a compact localizable string.

We should add a general step to the validation section that says for all properties with a known value type, issue a warning if the value does not match that type.

mattgarrish:
Closing this with #68.

Issue #62, opened 2019-09-16
"Processing the manifest": consider adopting the failure/error terminology


rdeltour:
The result of the current algo is either a canonical representation of the manifest or an early termination (which means "the manifest is not valid"). There are also cases where the algo issues a _warning_ but does not terminate.

Most of the other specs we reference (e.g. HTML, URL) instead use a _failure_/_error_ terminology. A failure is returned when the algorithm terminates early. An error is recoverable.

For instance, in the processing model for images [HTML defines _parse errors_](https://html.spec.whatwg.org/#concept-microsyntax-parse-error) as:

> A **parse error** for algorithms in this section indicates a non-fatal mismatch between input and requirements. User agents are encouraged to expose parse errors somehow.

The [URL Standard defines _validation error_](https://url.spec.whatwg.org/#validation-error) as:

> A **validation error** indicates a mismatch between input and valid input. User agents, especially conformance checkers, are encouraged to report them somewhere.
>
> Note: A validation error does not mean that the parser terminates. Termination of a parser is always stated explicitly, e.g., through a return statement.

I suggest we adopt the same terminology in our spec. The more alignment, the easier it is for spec ~~addicts~~ readers.

Also, incidentally, when the spec says:

> If the algorithm terminates early, the manifest is not valid.

I think it is worth noting that the opposite is not necessarily true. For instance when the required _type_ property is missing, a default value of `CreativeWork` will be assumed. I think this would be worth being explicitly clarified.

mattgarrish:
How would the following work for you:

>The steps for processing a manifest into its canonical representation are given by the following algorithm.
> 
> The following error types are used in this algorithm:
> 
>  - A validation error is a non-terminating error that occurs when the value of a property does match its expected input.
> - A fatal error is a terminating error that results, for example, when a manifest cannot be processed or does not match critical validity constraints.
> 
> User agents SHOULD expose both validation and fatal errors, but this specification does not prescribe the manner in which this is done.

We could then replace "error" with "fatal error" and "warning" with "validation error" as that's generally the way they are being used now (using dfn to link these).

(It's probably best not to delve into validity at all.)

rdeltour:
Sounds good to me. It doesn't reuse the _failure_ term used in the other specs I referenced earlier, but as long as we properly define these terms and point back to their definition when we use them, it's OK!

mattgarrish:
>  It doesn't reuse the _failure_ term used in the other specs I referenced earlier

[...]

PR #68, opened 2019-09-17
Updates to processing algorithm

### LINKED ISSUE #60: Handling of invalid values
We state to issue warnings when certain values are determined to be invalid, but that's dodging the issue of what a user agent has to do in these cases.

We can leave it entirely up to the user agent to determine what to do with invalid values, but we might want to look at the cases more closely to avoid implementation ambiguity.

Where the invalidity isn't significant, like dates, we might state, for example, that the reading system should not include the property in its internal representation. Likewise, if there's a default, as with reading progression, that should probably be substituted for the invalid value.

mattgarrish:
This PR fixes #60, #62 and #65 by making the following changes:

- it adds a validation check that properties have their expected value types
- it adds definitions for validation and fatal errors and updates algorithm to return failures
- it adds a statement that invalid values should be omitted (we don't need to say to remove the values anymore within the algorithm, so some checks have been modified accordingly)
- it adds a statement that errors should be presented

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/68.html" title="Last updated on Sep 18, 2019, 11:28 AM UTC (20ca22c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/68/fc0b219...20ca22c.html" title="Last updated on Sep 18, 2019, 11:28 AM UTC (20ca22c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2954,15 +2954,16 @@ enum ProgressionDirection {
 					<p>Perform data integrity checks on the following properties in <var>processed</var>:</p>
 					<ol style="list-style-type: lower-alpha;">
 						<li id="processing-checks-values">
-							<p>For each property with a known <a href="#properties-value-categories">value type</a>,
-								check that the actual value matches the expected value. If not, this is a <a>validation
-									error</a>.</p>
+							<p>(<a href="#properties-value-categories"></a>) For each property with a known <a
+									href="#properties-value-categories">value type</a>, check that the actual value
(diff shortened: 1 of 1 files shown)


mattgarrish:
I made a minor tweak to add a statement that the algorithm returns an object containing the canonical representation if successful, and also add a concluding step to return `processed`.

Otherwise, it seemed a bit odd to be "returning" failure for fatal errors, but letting the algorithm silently end if none occur.

PR #76, opened 2019-09-23
Move object definitions from appendix

### LINKED ISSUE #64: Custom Type Definition Appendix is confusing
`LinkedResource` is a type meant to be used within the JSON document, but `CreatorInfo` is not--as I understand it's use...and `LocalizableString` would actually create invalid JSON-LD if used in the document...
https://www.w3.org/TR/pub-manifest/#app-custom-types

We should either clarify their use ("internal representation only") or put them in separate sections or make them more clearly distinct "animals."

mattgarrish:
This is an attempt to resolve #64. It makes the following changes:

- object definitions are moved from the appendix to their corresponding value categories in 2.6.2.
- webidl dictionary definitions are made subsections of the PublicationManifest declaration
- CreatorInfo is changed to Entity to avoid the current proliferation of names for things - it is also explained that it is only a placeholder for Person or Organization
- Links (value category) is changed to Linked Resources per previous item
- the `type` property is removed from LocalizableString

Does this improve the clarity @BigBlueHat @iherman ?

### DIFF
--- a/index.html
+++ b/index.html
@@ -316,19 +316,19 @@ dictionary PublicationManifest {
              sequence&lt;DOMString>         accessibilityFeature;
              sequence&lt;DOMString>         accessibilityHazard;
              LocalizableString           accessibilitySummary;
-             sequence&lt;CreatorInfo>       artist;
-             sequence&lt;CreatorInfo>       author;
-             sequence&lt;CreatorInfo>       colorist;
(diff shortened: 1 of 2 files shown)


BigBlueHat:
Yep. Much clearer. Thanks, @mattgarrish!

PR #85, opened 2019-09-26
Schema improvements

### LINKED ISSUE #84: must inLanguage be an array?
I was trying out the JSON schema today, and got an error for

```JSON
"inLanguage": "en";
```

I suspect that others will be tripped up by this.

iherman:
- inLanguage should be a choice of a single bcp strings or an array thereof
- resource categorization properties should allow a single value instead of an array

fix #84


iherman:
I have just made the PR but did not have the time to test it (going to a two-days meeting soon). @dauwhe, would you mind running your test on the schema? Thx.

dauwhe:
Ivan, I downloaded the python version of jsonschema, and tried it on my document.

With the version of the schema in master, it ran fine, complaining that `inLanguage` should be an array.

With this version of the schema I got a large number of python errors, which seemed to relate to the `bcp.schema.json` file. I'm likely doing something wrong, but the very different behavior between the two branches is strange.

llemeurfr:
I don't have issues with this version, using ajv.

iherman:
Unfortunately, schemas are moving target. As far as I know, ajv is the most authoritative implementation around, so the fact that the schemas are o.k. with it is a good sign. Anyway, @dauwhe, can you send me the problems the python implementation have, to see if that can be taken care of.

This is why we clearly say that the json schemas are not normative...

mattgarrish:
>  I'm likely doing something wrong, but the very different behavior between the two branches is strange.

If you have this id in `publication.schema.json`:

>     "$id": "https://w3c.github.io/pub-manifest/schema/publication.schema.json",

Your validator is probably trying to grab the schemas from the server. That's why it would choke on the new file and still give the array error, as you have to temporarily make it a local reference:

>     "$id": "publication.schema.json",

mattgarrish:
Any reason we need the full URL in the `$id`?

I just went through this again adding the url regex and had to flip a whole bunch of files to get the schema to validate. I don't know if all validators use the id to find the source, but it's annoying if you have one that does.

iherman:
> Any reason we need the full URL in the $id?

To be honest I do not know. Let us do without and see what happens...

mattgarrish:
> To be honest I do not know. Let us do without and see what happens...

Okay, I just updated them. @dauwhe can you see if this makes validation work better for you, too?

mattgarrish:
FYI, I was looking through the spec and documentation and it's expected behaviour that when you specify an absolute URL in `$id` then files on the local system are ignored. It's in the documentation here: https://json-schema.org/understanding-json-schema/structuring.html#the-id-property

It's only best practice to use absolute URLs, though, so what we've done isn't forbidden. Maybe whenever the schemas stabilize we can switch back.

dauwhe:
I got this version to run in Python by:

1. Putting the schema folder on a local server and changing all the ids to something like `  "$id": "http://127.0.0.1/~cramerd/schema/url.schema.json"`

2. Using a much simpler regex for URLs. At least this jsonschema implementation wants you to double-escape everything, which makes it unusable in my opinion.

I'd say just merge this; it will make testing easier.

PR #86, opened 2019-09-26
Various fixes

mattgarrish:
This PR offers up a few changes:

- it adds conformsTo for declaring profile conformance (if accepted, schema to be updated after #85);
- it moves the value categories higher up in the manifest section so you encounter them before they start getting used;
- it adds a definition of "bounds" (see https://github.com/w3c/audiobooks/issues/26) and incorporates some prose from web pub that discussed it;
- it adds a definition of "profile" to the terminology so we're not just referencing the extensibility section;
- it updates the properties index

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/86.html" title="Last updated on Sep 30, 2019, 11:16 AM UTC (6107225)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/86/f45ef8f...6107225.html" title="Last updated on Sep 30, 2019, 11:16 AM UTC (6107225)">Diff</a>


iherman:
Actually, one more thing. We have now the extra example  to set a namespace in an extra context. We will, soon, have direction added there, too. We should make it clear that all these specification can be added in one object instead of three, and the json schema should be extended to allow for all these, too. (I would think that, if possible, that part of the json schema should be open ended, rather that specifying in details. I am not sure from the top of my head, how to do that in schema.)

(Sorry, I am on a meeting, I can't do more at the moment...)

mattgarrish:
>     * conformsTo shoukd be added to the json schema, too

Right, I was going to do that after you merge #85 to avoid any conflicts, but maybe there's no overlay.



>     * there is no mapping of conformsTo to schema, but maybe we should make it explicit that this is a dcterms term, and add a reference to dcmi instead in the table

Ya, I wonder if we can make the last column in the table a bit more flexible so it's not always about a schema.org mapping? In this case, we just swap it to "DCMI Mapping" or something like that.



>     * shouldn't conformsTo be added to the WebIDL?

It should be there after the type property (second item).

> We should make it clear that all these specification can be added in one object instead of three

Right, I can modify the example to show the language declaration in the object, as well.

mattgarrish:
> I would think that, if possible, that part of the json schema should be open ended, rather that specifying in details.

Ya, I'm not seeing how we can enforce anything via schema without clamping down on what people can write into the context. It's a pretty rudimentary language.

Beyond saying everything else you declare has to follow the contexts, any additional verification probably has to be done by other means.

dauwhe:
> Beyond saying everything else you declare has to follow the contexts, any additional verification probably has to be done by other means.

I still have hopes of pub manifest not actively conflicting with WAM. Right now that seems to be true; I can add a few WAM-specific members to a pub manifest and have it valid to this schema and not alarm the WAM validators. Haven't tried to get a working PWA like this yet, though.

So I hope we don't lock down the manifest, prevent any unknown elements, etc.

mattgarrish:
> So I hope we don't lock down the manifest, prevent any unknown elements, etc.

No, this is strictly about what we expect in the ```@context``` declaration, in what order, etc.

As far as properties in the manifest, the whole idea is to be as open as we can to extensibility. That's why I found the use of ```type``` for profile identification worrisome.

iherman:
If and when merged, https://github.com/w3c/pub-vocab/pull/8 should also be merged (adding conformsTo to the context file).

iherman:
Shouldn't the schema for `conformsTo` be:

```
   "conformsTo" : {
            "oneOf": [
                {
                    "$ref": "url.schema.json"
                },
                {
                    "type": "array",
                    "items": {
                        "$ref": "url.schema.json"
                    }        
                }
            ]
       }
```

iherman:
>> I would think that, if possible, that part of the json schema should be open ended, rather that specifying in details.
>
> Ya, I'm not seeing how we can enforce anything via schema without clamping down on what people can write into the context. It's a pretty rudimentary language.
> 
> Beyond saying everything else you declare has to follow the contexts, any additional verification probably has to be done by other means.

Per the schema specification[1] any term can be added to an object, and the validation is relevant for those terms that are explicitly specified. Which is fine: in our spec we do provide a schema for `@context` covering `@language`, possibly other terms later that we would explicitly refer to, but anything else should be covered by the JSON-LD specification. (Maybe that should be explicit in the text.)

[1] https://json-schema.org/understanding-json-schema/reference/object.html

mattgarrish:
> Shouldn't the schema for `conformsTo` be:

Oops, yes, just updated.

mattgarrish:
> Per the schema specification[1] any term can be added to an object, and the validation is relevant for those terms that are explicitly specified.

Right, but there's no way to issue an error that this is wrong, at least that I can spot:

```
"@context" : [
    "https://schema.org",
    "https://www.w3.org/ns/pub-context",
    { "language" : "en" },
    { "language" : "fr" }
]
```

It means there are potentially confusing overlaps we can't warn about through the schema alone.

I also noticed that our current language definition doesn't validate. If you add a language that doesn't match the language definition, the object just gets treated as though it's some other unspecified declaration.

If we set `"additionalItems" : false` to the context, then the language gets validated,  but it also becomes invalid to add anything else to the context but the two contexts and an object. (But inside that object you can add whatever you want.)

iherman:
> Right, but there's no way to issue an error that this is wrong, at least that I can spot:
>
>```
>"@context" : [
>   "https://schema.org",
>    "https://www.w3.org/ns/pub-context",
>    { "language" : "en" },
>    { "language" : "fr" }
> ]
> ```

Yep. And, actually, this is perfectly valid JSON-LD (and French will win:-). It may be possible to restrict the content of that outer array so that it contains the first two or one more extra object (I presume by using "oneOf" and two arrays with explicit contents) but I do not think we should do that. That may be a step too far in restricting the "shape".

> I also noticed that our current language definition doesn't validate. If you add a language that doesn't match the language definition, the object just gets treated as though it's some other unspecified declaration.

That _is_ strange. But if limiting the object is the only way around it, I guess we should let this one go.

Luckily the json schema is _not_ normative, and we make it clear in the document...:-)

mattgarrish:
>  And, actually, this is perfectly valid JSON-LD

Yes, but what does it mean as far as the processing algorithm goes? What do we say there about calculating these values? Does it need to be noted that you have to follow json-ld processing rules for the context to get the global language and direction?

iherman:
>> And, actually, this is perfectly valid JSON-LD
>
> Yes, but what does it mean as far as the processing algorithm goes? What do we say there about calculating these values? Does it need to be noted that you have to follow json-ld processing rules for the context to get the global language and direction?

God forbid :-)

We understand a very limited number of those top level json-ld statements in the separate contexts. At the moment, this is limited to `@language` and it will also include `@direction` in an upcoming PR. So I guess what we can say is that the processing follows all objects in the array (after the url to our own context) and looks for `@langauge` (and `@direction`). The latest entry in the array wins; everything is ignored for our processing rules.

Or... we do restrict, via the json schema, that our context may consist of two or three entries with the third entry containing `@langauge` and `@direction`. But I am a bit afraid that this is too limiting if someone wants to do further, funky things with JSON-LD contexts.

mattgarrish:
> Or... we do restrict, via the json schema, that our context may consist of two or three entries with the third entry containing `@language` and `@direction`

Ya, I don't really like the idea of restricting the authoring side. It's probably not even that likely to come up, and given that it's valid JSON-LD I guess you do mess up your definitions at your own peril.

I guess picking the last language/direction specified in the context is probably the closest to reality we're likely to get from a non-JSON-LD processor, but how do we account for JSON-LD processing? Do we distinguish the selection process based on capabilities? I think this is the only case where it matters what kind of processor is doing the ingestion.

iherman:
>  I think this is the only case where it matters what kind of processor is doing the ingestion.

Nope, we are fine. JSON-LD does exactly the same: if there are identical term definitions in consecutive contexts, the last one wins. Look at the [playground example](https://tinyurl.com/yyt3cy7g): if you look at the expanded view (which means all definitions in the contexts are handled yielding something fairly closed to (actually a subset of) our processed data, you get the second language set.

mattgarrish:
> if there are identical term definitions in consecutive contexts, the last one wins

Unless you `@protected` the first one, for example. Then our algorithm will produce a different result from an actual json-ld processor (which I understand would probably abort processing).

But it's probably rare enough it's not going to matter. :)

Is there any formal terminology for talking about a "top-level" declaration, though? We obviously don't expect "fr" to result from the following:

```
"@context" : [
    "https://schema.org",
    "https://www.w3.org/ns/pub-context",
    { 
        "language" : "en",
        "term":  {
            "@context" : {
                "language" : "fr"
            }
        }
    }
]
```

mattgarrish:
> Is there any formal terminology for talking about a "top-level" declaration

Would the following work:

> Set `language` to the [default language](https://w3c.github.io/json-ld-syntax/#dfn-default-language) [[!json-ld]] of `manifest` (i.e., the last non-scoped `language` declaration in `manifest[@context]`).

That's my best attempt so far to dodge being committal about how it has to be compiled.

iherman:
> Unless you @protected the first one, for example. Then our algorithm will produce a different result from an actual json-ld processor (which I understand would probably abort processing).

Yes, and a lot of other things may happen if the various JSON-LD 1.1 features are used. Actually, it can also happen for simpler structures: there is no protection in the schema.org context, meaning that the author may overwrite any term definition in his/her context, turn a schema.org/name into something very different. But I do not think we want to go there. A JSON-LD 1.1 expert person ought to know that many things may be changed.

>> Set language to the default language [[!json-ld]] of manifest (i.e., the last non-scoped language declaration in manifest[@context]).
>
> That's my best attempt so far to dodge being committal about how it has to be compiled.

That sounds about right from JSON-LD point of view, but it is a bit scary. This means the reader of the processing section should understand JSON-LD 1.1 terminologies, and we are aiming mostly at people who do not (and do not want to) know about JSON-LD (of any version). What about something like:

> Set language to the of the last apperance of the name `"@language"` of the json objects (if any) appearing in the `"@context"` array. 

That would rule out any scoped context.

The term "name" is used in [ECMA JSON](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf), which is a little bit misleading for us. I guess we use the term, ehem, `"term"`, so we could safely use that. (Or say "name (term)" to avoid ambiguity.)

mattgarrish:
I've made a few more updates. I made it so that a non well-formed language tag won't propagate out from the global declaration. I also stripped the language object from the schema since it doesn't do anything and is probably not what we want even if it did.

iherman:
>  I also stripped the language object from the schema since it doesn't do anything and is probably not what we want even if it did.

I removed the unnecessary comma from the context part of the schema (which would make it invalid JSON...)

Otherwise: in my view we can ship it!

PR #88, opened 2019-09-27
change format=uri and iri to regexes

mattgarrish:
I took the regex from here: https://stackoverflow.com/a/190405

I modified the start for https? and add A-Z A-F to the regex as I couldn't find a way to specify case insensitive matching in schemas.

Not heavily tested, but seems to work for general cases.


iherman:
Wow. That is quite a regex...

Your next job will be to explain in details what the regex does:-)

mattgarrish:
> Your next job will be to explain in details what the regex does:-)

It matches IRIs, of course. Explanation over... ;)

PR #89, opened 2019-09-30
Update publication.schema.json

iherman:
I believe we have to put `"additionalItems":true`, otherwise the extra objects will not be allowed. We _could_ refine this by sayint somethings like

```
"additionalItems" : {
    "oneOf" : ...
```

for url-s or strings, but I would think we may stop here.

(I am at the airport here, soon to board, so I cannot check/go into details here...)

### DIFF
--- a/schema/publication.schema.json
+++ b/schema/publication.schema.json
@@ -14,6 +14,7 @@
                     "const": "https://www.w3.org/ns/pub-context"
                 }
             ],
+            "additionalItems" : true,
             "uniqueItems": true
         },
         "type": {


mattgarrish:
I believe it is true by default when not specified:

> Omitting this keyword has the same behavior as an empty schema.

I take that to mean it's equivalent to `"additionalItems": {}` when not set, which allows additional items without any restriction on their type. I haven't run into any problems adding items, at least.

But no harm in being explicit, either.

mattgarrish:
> by saying somethings like ... for url-s or strings, but I would think we may stop here.

Seems reasonable. Would you ever have an array inside the array? Other than that, it doesn't look like the other data types would be meaningful inside a context array.

iherman:
This should have been merged into the other PR before merging that one, but was missed...

I have made a manual change directly on the master branch in https://github.com/w3c/pub-manifest/commit/40a86ee1f50084a4899026143cf86826705f7879.

mattgarrish:
Oops, I wasn't paying attention to the destination branch... :)

PR #92, opened 2019-10-02
Settting the base direction

### LINKED ISSUE #75: Text base direction again — but now with a solution:-)
The issue of base direction has been plagued us for a while (see also [issue 354 on wpub](https://github.com/w3c/wpub/issues/354)), but we may have a proper solution for it now. There has been discussions [elsewhere](https://w3c.github.io/rdf-dir-literal/) to look for a solution, a (failed) attempt to revive the discussion in RDF land and, finally, a [breakout session at TPAC](https://www.w3.org/2019/09/18-rdf-dir-literal-minutes.html). This led the JSON-LD WG to re-open the [issue of adding base direction to JSON-LD 1.1](https://github.com/w3c/json-ld-syntax/issues/11). On its F2F meeting at TPAC, the JSON-LD WG has accepted a series of resolutions, see the minutes of [Thursday](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-09-19-json-ld) and [Friday](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-20-19-json-ld) for the details.

The essence is: JSON-LD 1.1 will introduce a new keyword `@direction` that can be used, essentially, the same way as `@language`: can be part of the context to denote a global (ie, default) value, and can also be used as part of an individual literal. In our terminology, a `LocalizableString` in our manifest can have a directional value, if needed, just the same way as we handle language tags. 

_What I propose is to make changes on the manifest specification to adopt this feature._ What it would mean for the manifest is:

- the current `direction` term disappears and is replaced by a `@direction` keyword usage alongside `@language` (see [2.5.1](https://w3c.github.io/pub-manifest/#manifest-lang-dir-global))
- [item specific settings (in 2.5.2)](https://w3c.github.io/pub-manifest/#manifest-lang-dir-local) will include `@direction`
- The [definition of `LocalizableString`](https://w3c.github.io/pub-manifest/#LocalizableString) will be updated
- Some additions to the context file to empower all this.

The result, I believe, would be a _much_ cleaner format for our manifest, and we can but this issue at rest at last.

There is one caveat, though. This is a _JSON-LD 1.1 feature_, i.e., we become dependent on JSON-LD 1.1. This, by itself, is not a problem, JSON-LD 1.1 is slightly ahead of use in its advance towards a Rec. However, we have to be careful to use this feature in a way that it does not upset JSON-LD 1.0 processors that are supposed to simply ignore an unknown keyword. Without going into details what this means that we should _not_ create the `direction` alias for `@direction`, as we do for, e.g., `@language` and `@value`. As a consequence, _I would propose that we remove the usage of these aliases, and we use `@langauge` and `@value`; otherwise it looks very inconsistent_.

As soon as the JSON-LD 1.1 editor's draft includes the new feature, I will put in a PR to adopt these changes in the manifest document. We can decide, through the PR, whether we agree with these changes.

(This should supersede, ie, close, issue #39, and also make the discussions in https://github.com/w3c/wpub/issues/354 moot. Finally, it should close the gap in the i18n review in #38.)

Cc: @r12a @aphillips @mattgarrish @wareid @BigBlueHat @laudrain @llemeurfr @GarthConboy @rdeltour @dauwhe

iherman:
I have incorporated the usage of the upcoming JSON-LD 1.1 `@direction` (see the adjacent [PR No. 276](https://github.com/w3c/json-ld-syntax/pull/276) in the JSON-LD WG).

To facilitate review, here are the points were the essential changes happened:

- 2.4.4.1, adding direction to Localizable Strings, plus examples
- 2.6 on global and local setting of language & direction
- Modifications in section 4, steps (6), (7c), (9d)
- Added Appendix E

See also #75 for the background and the writeup from the TPAC discussions

Cc: @aphillips, @r12a, I would particularly welcome your check on the examples in 2.4.4.1 and 2.6

Cc: @dauwhe @laudrain @gkellogg

---

Fix #75

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/92.html" title="Last updated on Oct 4, 2019, 2:47 PM UTC (060d46b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/92/35a23c9...060d46b.html" title="Last updated on Oct 4, 2019, 2:47 PM UTC (060d46b)">Diff</a>

### DIFF
--- a/common/js/biblio.js
+++ b/common/js/biblio.js
@@ -103,6 +103,15 @@ var biblio = {
       ],
       "date": "2017-12-01"
     },
+    "json-ld10": {
+      "title": "JSON-LD 1.0",
+      "authors": [
+        "Manu Sporny", "Gregg Kellogg", "Markus Langhaler"
+      ],
+      "status": "W3C Recommendation",
+      "date": "2014-01-16",
+      "href": "https://www.w3.org/TR/2014/REC-json-ld-20140116/"
+    },
     "schema.org": {
         "title": "Schema.org",
         "href": "https://schema.org"
--- a/index.html
+++ b/index.html
@@ -528,12 +528,24 @@ dictionary LocalizableString {
 										</code>
 									</td>
 									<td>The base direction of the value. OPTIONAL.</td>
-									<td>"ltr" or "rtl"</td>
+									<td><code>ltr</code> or <code>rtl</code></td>
 									<td><a href="#value-literal">Literal</a></td>
 									<td>(None)</td>
 								</tr>
 							</tbody>
 						</table>
+
+						<p>The meaning of the base direction values are:</p>
+
+						<ul data-dfn-for="TextDirection">
+							<li><code><dfn>ltr</dfn></code>: indicates that the textual value is explicitly directionally set to left-to-right text;</li>
+							<li><code><dfn>rtl</dfn></code>: indicates that the textual value is explicitly directionally set to right-to-left text;</li>
+						</ul>
+
+						<p>
+							A missing base direction value means that that the textual value is explicitly directionally set to the direction of the first character with a strong directionality, following the rules of the Unicode Bidirectional Algorithm&#160;[[!bidi]].</li>
+						</p>	
+
 						<aside class="example" title="Set the language of a string">
 						<pre>{
     "value"     : "孔子",
@@ -549,11 +561,11 @@ dictionary LocalizableString {
 					</aside>
 					<div class=note>
 						<p>
-							If the base direction value were not set in the last example, the text would be displayed, following the Unicode Bidirectional Algorithm&nbsp;[[bidi]], as
+							If the base direction value were not set in the last example, the text would be displayed, following the Unicode Bidirectional Algorithm&nbsp;[[bidi]] and due to the presence of a Latin character starting the string, as:
 						</p>
 						<p dir="ltr">HTML היא שפת סימון.</p>		
 						<p>
-							due to the presence of a, in this case, Latin character starting the sentence. However, that would be incorrect. The extra <var>direction</var> value is necessary to yield the right display, namely:
+							However, that would be incorrect. The extra <var>direction</var> value is necessary to yield the right display, namely:
 						</p>
 
 						<p dir="rtl">HTML היא שפת סימון.</p> 
@@ -1148,13 +1160,13 @@ dictionary LocalizableString {
 
 					<p>The global language and base direction declaration, when present, MUST follow the <a>publication context</a>.</p>
 
-					<p>Default values are not specified for the global language and base direction declarations.</p>
+					<p>Default values are not specified for the global language. The default value for the global direction is <code>null</code>.</p>
 				</section>
 
 				<section id="manifest-lang-dir-local">
 					<h4>Item-Specific Declarations</h4>
 
-					<p>It is possible to set the language locally for any natural language value in the manifest using a
+					<p>It is possible to set the language or a base direction locally for any natural language value in the manifest using a
 							<a href="#LocalizableString">localizable string</a>:</p>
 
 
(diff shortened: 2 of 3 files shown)


mattgarrish:
Generally looks fine to me, but weren't we going to leave the json-ld references as latest links? I understand that direction is new, but didn't we also expect 1.1 to be on the same track to standardization as this doc so we wouldn't end up referencing 1.0?

iherman:
Yes, I guess we can switch 'back' to json-ld reference at Proposed Rec; by then 1.1 may be a Rec. I would propose to leave things as in this PR for now, to make the dependency explicit, and come back on this later. (It will be an editorial change.)

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-10-07-pwg#section1).

- `RESOLVED:  ivan to merge direction PR`
Directionality<br/>
Wendy Reid: <a href="https://github.com/w3c/pub-manifest/issues/75">https://github.com/w3c/pub-manifest/issues/75</a><br/>
Wendy Reid: first issue is about directionality<br/>
Ivan Herman: See Pull request for directionality<br/>
Ivan Herman: I have already reported that this whole direction issue now has a clear solution<br/>
… I’ve made a PR<br/>
… you can do direction much like you can do language<br/>
… with the same syntax, either globally or for a given member<br/>
… so the PR fixes all the bits<br/>
… there is one issue which I put into the document<br/>
… having direction in the system means that we have a dependency on JSON LD 1.1<br/>
… which is running in parallel with our spec; they plan CR at the same time, they have implementations.<br/>
… so they may be ahead of us<br/>
… but if you want to run a manifest through a JSON-LD processor that only supports 1.0 you will get errors<br/>
… so I created a separate context file that doesn’t have direction<br/>
… and made an appendix in the doc<br/>
… so you can use this context if you don’t need direction, and need to use a JSON-LD 1.0 processor<br/>
… so the question to the group is that it’s OK to keep this as an appendix in the document<br/>
… or put it somewhere else, or not do it at all and depend on JSON-LD 1.1<br/>
… so I want help from the group in deciding what to do<br/>
Wendy Reid: any questions?<br/>
Brady Duga: I don’t have strong feelings, but it does feel strange to put in the spec itself<br/>
… most things that process this won’t be JSON-LD processors<br/>
… but I’m not against it<br/>
… I think it should be in a read me or impl guide<br/>
Matt Garrish: I’m concerned about how this might be interpreted<br/>
… changing our context which is required seems weird<br/>
… it could add confusion to have an alternate context<br/>
Ivan Herman: I put it there without being convinced it’s a good idea<br/>
… I’m fine removing it<br/>
… so where would this information go?<br/>
… we could give it to the testing people and have this added to information about testing<br/>
Matt Garrish: I would put in our wiki for implementation details, so it’s discoverable<br/>
Wendy Reid: a wiki is good<br/>
Dave Cramer: This seems like it would be a helpful note in the spec itself, if this language property is used, mention that JSON-LD 1.1 is required<br/>
… it’s a good use of non-normative text<br/>
Ivan Herman: any other comments?<br/>
… one comment… for those of you who have been looking at this<br/>
… if you look at this example<br/>
… it is a good example there for real cases where base direction is necessary<br/>
… some people were wondering if this stuff was necessary<br/>
… We found a book with a title in arabic and english<br/>
Ivan Herman: so it’s good to have a real-world example<br/>
Matt Garrish: I think it’s useful to have direction; we have it in EPUB.<br/>
Ivan Herman: is there a consensus to merge?<br/>
Matt Garrish: +1<br/>
Wendy Reid: +1<br/>
Brady Duga: +1<br/>
Dave Cramer: +1<br/>
Gregorio Pellegrino: +1<br/>
Ivan Herman: we should also pledge to never talk about direction again :)<br/>
Resolution #2: ivan to merge direction PR<br/>
</details>

r12a:
> Cc: @aphillips, @r12a, I would particularly welcome your check on the examples in 2.4.4.1 and 2.6

They look good to me.

Issue #95, opened 2019-10-03
constraints on conformsTo?


iherman:
The [processing steps](https://w3c.github.io/pub-manifest/#manifest-processing) include two fairly stringent requirements on the value of `conformsTo` in step (5):

- At least on value of `conformsTo` should refer to a 'recognizable profile' (and this is a possibly fatal error), although we do not talk about what 'recognizable profile' means. We should at least give some details on that in the text, but that can easily lead to some sort of a registry service, for example. I am not sure we should go down that route.
- Only one recognizable profile should be listed. Do we want to be that restrictive? Ie, even if we have recognizable profiles like, say, audiobook or manga, is it out of the question that a manifest would bear _both_ profiles? Is this a restriction we want to build into the manifest spec?

My initial gut feeling is to drop both of these requirements, because it is difficult to check.

mattgarrish:
>     * Only one recognizable profile should be listed. Do we want to be that restrictive?

The processing step doesn't say to ignore others:

> The user agent can select which to give precedence to.

The user agent has to determine what profile(s) it recognizes as it directly affects how it processes the extension steps later. It may also affect how it presents the content - whether as an audio interface or an image one, for example.

The part about being a validation error should go. It may be problematic to have more than one listed, but I suppose that would come out in validation. What if instead it says:

> If manifest["conformsTo"] includes the URLs of more than one recognized profile, the user agent can select which to give precedence to in the case of conflicts.



> although we do not talk about what 'recognizable profile' means

One that it is "capable of processing and/or rendering"?

iherman:
> The part about being a validation error should go. It may be problematic to have more than one listed, but I suppose that would come out in validation. What if instead it says:
>
> "If manifest["conformsTo"] includes the URLs of more than one recognized profile, the user agent can select which to give precedence to in the case of conflicts."

👍  

>> although we do not talk about what 'recognizable profile' means
>
> One that it is "capable of processing and/or rendering"?

Yep.

llemeurfr:
Letting the user agent decide what to do with ambiguous information is a breach in interoperability, isn’t it? I don’t like it.

mattgarrish:
> Letting the user agent decide what to do with ambiguous information is a breach in interoperability

First in the array wins?

llemeurfr:
> Letting the user agent decide what to do with ambiguous information is a breach in interoperability
> 
> First in the array wins?
> 

This is quite logical indeed.

mattgarrish:
[...]

Issue #94, opened 2019-10-03
Ordering of processing steps v.a.v. conformsTo


iherman:
At the moment, in the [Processing Steps sequence](https://w3c.github.io/pub-manifest/#manifest-processing):

- step (5) checks the values of profile conformance, ie, the values of `conformsTo`
- step (7.e) turns possibly relative URL-s into absolute ones.

The value of `conformsTo` is set to be URL-s, i.e., they can also be relative ones (even if not frequently). Either we have to say that they are always absolute URL-s (which must then be checked), or the order should be reversed. 

I believe if the current step (5) goes after (7) then I believe we are fine.

mattgarrish:
> i.e., they can also be relative ones

Wouldn't that make it a profile local to one file, that changes wherever the file is moved to? What would be an example?

I'd think we either should require absolute URLs or we should be treating conformsTo as a literal that takes a URL as its value. That way if someone were to use a relative URL, which sounds like a keyword in disguise, that relative URL won't get turned into something it isn't later.

iherman:
> I'd think we either should require absolute URLs or we should be treating `conformsTo` as a literal that takes a URL as its value. 

I do not have a problem with that either.

mattgarrish:
Closing this with the merge of #96

PR #96, opened 2019-10-03
Update conformsTo

mattgarrish:
This PR updates conformsTo and its processing per the issues raised in #94 and #95.

It makes the value of conformsTo a literal so that URL processing doesn't affect the values, and clarifies the presence of multiple profiles.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/96.html" title="Last updated on Oct 3, 2019, 5:31 PM UTC (a55d37d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/96/41b0957...a55d37d.html" title="Last updated on Oct 3, 2019, 5:31 PM UTC (a55d37d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -140,7 +140,7 @@
 					</dd>
 
 					<dt>
-						<dfn>Profile</dfn>
+						<dfn data-lt="profiles|profile(s)">Profile</dfn>
 					</dt>
 					<dd>
 						<p>Profiles are publication formats (e.g., audiobooks) that use the <a>manifest</a> format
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1264,7 +1263,7 @@ <h3>Profile Conformance</h3>
 							</td>
 							<td>URL of the profile.</td>
 							<td>A valid URL string&#160;[[!url]].</td>
```
iherman: Maybe, to emphasize, we could say "Absolute URL of the profile"

mattgarrish: For the value, I can just change "valid URL string" to "absolute-URL-with-fragment string".

Issue #101, opened 2019-10-07
Lack of conformsTo too draconian?


mattgarrish:
Right now, it's a fatal error if you don't specify `@context` or identify the profile in `conformsTo`.

Without a common base, like we had with web publications, there isn't a default profile that we can recommend user agents fall back on to try and open a publication.

So the question is, can people live with the fatal error, or should we try and find some compromise wording to allow a user agent to attempt to open an unknown manifest using a profile it does know how to process?

iherman:
... or we fall back on requiring the https://www.w3.org/TR/pub-manifest, or setting this one as a default.

mattgarrish:
> https://www.w3.org/TR/pub-manifest, or setting this one as a default

That would give it a value, but what does it mean for processing? There are no behaviours associated with a generic pub manifest.

iherman:
It could mean something like no specific behaviour is defined for that manifest (apart from the processing steps defined in the spec), and the URL just says that the manifest is meant to be a valid manifest per the pub-manifest specification.

Obviously not very useful. But it makes testing, etc, more straightforward; it may also allow unforeseen or experimental readers to develop with some local knowledge without the need to change the manifest.

(To be clear: I did not think it through 100%, so there may be fatal caveats.)

mattgarrish:
> It could mean something like no specific behaviour is defined for that manifest

In theory, yes, but what does a dedicated audiobook player presented with this manifest do, for example? Won't it have to try and open the publication as an audiobook if it doesn't barf on it?

If it was just the web we were talking about it probably wouldn't matter, you could process the manifest and just leave whatever linked to the manifest alone as a web page. But for distributed content there's going to be some expectation of what the package contains. I don't think we can ever capture what that expectation is, though.

It seems like the user agent could opt to either: 1) issue a fatal error and not render the content; or 2) issue a validation error and continue processing based on whatever it is capable of rendering.

iherman:
> In theory, yes, but what does a dedicated audiobook player presented with this manifest do, for example? Won't it have to try and open the publication as an audiobook if it doesn't barf on it?

To be honest, I do not really know but, clearly, if somebody creates an audiobook, than he/she should definitely set the right pointer.

Again, I would regard such a pointer (ie, to the pub-manifest) more as some sort of a testing/development conformance profile, without specifying it further. Which, I guess, means that my option would be:

> 2) issue a validation error and continue processing based on whatever it is capable of rendering.

[...]

Issue #98, opened 2019-10-05
WebIDL, yea or nay?


rdeltour:
The Publication Manifest specification [relies on WebIDL](https://w3c.github.io/pub-manifest/#webidl) to define the internal data structure of manifests.

One of the reasons we chose WebIDL is that we were influenced by its use in [Web App Manifest](https://w3c.github.io/manifest/#idl-index). Now, it seems that the Web App Manifest folks are reconsidering that, see the entire discussion in w3c/manifest#611. (Thanks @marcoscaceres 🎂 for hinting at this discussion at TPAC).

The TAG also commented on our use of WebIDL, to which [@iherman replied](https://github.com/w3ctag/design-reviews/issues/344#issuecomment-491025385).

Are we sure that WebIDL is the right approach and won't create more future issues than it solves? (I’m not an expert in WebIDL myself, and still need to digest the issues raised for Web App Manifest).

In the context of the TAG issue "[the proliferation of manifests at W3C](https://github.com/w3ctag/design-reviews/issues/423)" by @tantek, I believe that things like how to describe data structures is typically worth being looked into, and see if these specs can or need to adopt a unified approach.

I know it’s late in our editing process, and we’ve had lengthy discussions on using WebIDL, and even more time spent on actual editing. But we'd better make sure we're on the right track before moving to CR.

iherman:
My answer [that you quoted](https://github.com/w3ctag/design-reviews/issues/344#issuecomment-491025385) still stands. Although the subsequent discussion in that thread was diverted into issues about the _usage_ of WebIDL and not whether to use WebIDL or not (which also led to substantive changes), the remark I made there:

> All that being said, I understand and share your unease about the usage of WebIDL; I think we would be happy to consider an alternative. We just did not find any...

still stands.

Although I admit I only had a cursory look https://github.com/w3c/manifest/issues/611 (it is a _very_ long thread) I did not see any alternative emerging either. At some point I raised the (half serious) idea of using TypeScript for the same purpose; I guess it would be perfectly readable and concise (I actually did, for my own learning, a [version](https://github.com/iherman/PubManifest/blob/master/src/manifest.ts) of the data structure definition), but we also agreed that it would not be wise to bind to an existing programming language.

[...]

PR #103, opened 2019-10-08
update manifest processing to use infra types

mattgarrish:
Opening this as a PR to get feedback, as I'm going cross-eyed working on this section. Beyond just switching terminology, I also tried to tighten up the processing to account for potentially other bits of invalid input, to simplify the duplication of language/direction steps, and to consolidate a number of validation checks that were all dependent on recursion.

The WebIDL section is moved into the processing section, and to de-emphasize the technology in the interim I changed the title to "Data Visualization" (but other ideas welcome).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/103.html" title="Last updated on Oct 13, 2019, 1:11 AM UTC (bf40720)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/103/cd6bbd6...bf40720.html" title="Last updated on Oct 13, 2019, 1:11 AM UTC (bf40720)">Diff</a>


index.html
```diff
@@ -100,6 +100,8 @@ <h3>Scope</h3>
 			<section id="terminology">
 				<h3>Terminology</h3>
 
+				<p>This specification depends on the Infra Standard [[!infra]].</p>
+
```
iherman: I am not sure it is appropriate to refer to the infra document normatively. I will have to check with the director, though, but there are only a few what wg documents (html, fetch, probably dom) that can be referred to normatively.

index.html
```diff
@@ -2992,174 +2882,221 @@ <h3>Other Discovery Methods</h3>
...
-					</details>
-				</li>
+					<li id="processing-conformance">
+						<p>(<a href="#profile-conformance"></a>) Determine the <a>profile(s)</a> the manifest conforms
+							to as follows:</p>
+						<ol>
+							<li>If <var>manifest["conformsTo"]</var> is not set, this is a <a>fatal error</a>. Return
+								failure.</li>
+							<li>If <var>manifest["conformsTo"]</var> does not include a URL of a profile it is capable
+								of processing and/or rendering, this is a <a>fatal error</a>. Return failure.</li>
```
iherman: It may be better to say "the user agent" instead of "it"...

index.html
```diff
@@ -2992,174 +2882,221 @@ <h3>Other Discovery Methods</h3>
...
+									<var>context["language"]</var>;</li>
+							<li>if <var>dir</var> is an empty <a href="https://infra.spec.whatwg.org/#string">string</a>
+								and <var>context</var> is a <a href="https://infra.spec.whatwg.org/#ordered-map">map</a>
+								that contains a <var>direction</var>
+								<a href="https://infra.spec.whatwg.org/#map-key">key</a>, set <var>dir</var> to
+									<var>context["direction"]</var>;</li>
+							<li>if both <var>lang</var> and <var>dir</var> are not empty <a
+									href="https://infra.spec.whatwg.org/#string">strings</a>, then <a
+									href="https://infra.spec.whatwg.org/#iteration-break">break</a>.</li>
+						</ol>
```
iherman: I am not sure that is correct. If the contexts are:

```json
[{ "language": "en", "direction":"ltr"}, {"language":"ar"}]
```

then the global language will be set to `en`, because it is only getting at that point that _both_ language and dir are not empty, i.e., a break would happen.

I have the impression that it is cleaner to (conceptually) run the cycles for language and for direction independently of one another, and not trying to merge them.

index.html
```diff
@@ -2992,174 +2882,221 @@ <h3>Other Discovery Methods</h3>
...
-										<code>Person</code>, but, for the sake of convenience, only their name has to be
-									specified. For example:</p>
-								<pre class="example">{
+								</details>
+							</li>
+
+							<li id="processing-normalize-entities">
+								<p>(<a href="#value-entity"></a>) If <var>term</var> expects an <a href="#value-entity"
+										>entity</a>:</p>
+								<ol>
```
iherman: Strictly speaking, this is incorrect. Per our definition, e.g., `author` _never_ expects and entity; it expects an _array of entities_. I presume that what we have to put there in the 'If' clause.

mattgarrish:
One question: should the webidl section also be informative?

We're not actually requiring webidl, from the processing, so it's more an example of the final data than a usable schema.

rdeltour:
> One question: should the webidl section also be informative?

You beat me to it, I was writing my review 😊: yes in its current form I think it should be informative.

index.html
```diff
@@ -3443,155 +3470,286 @@ <h2>Processing a Manifest</h2>
...
+											</li>
+										</ul>
+									</li>
+
+									<li id="processing-checks-length">
+										<p>(<a href="#LinkedResource"></a>) If <var>processed[term]</var> expects type
+												<a>LinkedResource</a> and its <var>termValue["length"]</var>
+											<a href="https://infra.spec.whatwg.org/#map-key">key</a> is set and is not a
+											valid number, this is a <a>validation error</a>.</p>
+									</li>
```
iherman: Shouldn't it say 'non-negative number'?

mattgarrish:
> If the idea is that we want a simple and terse way to _represent_ the data structure, is it a matter of defining a notation for Infra types?

Yes, that would be ideal, if it can happen in time with our charter. Given our current options, however, webidl has proven the most stable and understandable way of expressing this info.

But I agree with webidl being informative if we keep it. It's not expressing the actual final representation.

index.html
```diff
@@ -3443,155 +3470,286 @@ <h2>Processing a Manifest</h2>
...
+					</li>
+
+					<li id="processing-end">
+						<p>Return <var>processed</var>.</p>
+					</li>
+				</ol>
+			</section>
+
+			<section id="internal-rep-visualization">
+				<h3>Data Visualization</h3>
```
iherman: I would think this section should be non-normative.

I am not sure about the title either. Just some alternative that comes to my mind:

- "Data structure overview"
- "Overview of the internal data structure"

"Visualization" triggers the association with graphics in my mind...

index.html
```diff
@@ -3443,155 +3470,286 @@ <h2>Processing a Manifest</h2>
...
+											<li>
+												<p>otherwise, if <var>url</var> is not a valid URL&#160;[[!url]], this
+													is a <a>validation error</a>.</p>
+											</li>
+										</ul>
+									</li>
+
+									<li id="processing-checks-length">
+										<p>(<a href="#LinkedResource"></a>) If <var>processed[term]</var> expects type
+												<a>LinkedResource</a> and its <var>termValue["length"]</var>
```
iherman: This links to the WebIDL dictionary, which is problematic. We do not know if WebIDL will survive and, by now, the term "dictionary" has become banned:-)

It should probably link too 2.3.4.3., ie, `#value-linked-resource` instead.

iherman:
Actually, I may revise my earlier comment in another thread, and would opt to move the WebIDL into an (informative) appendix. If we want to change later by either adding other formalism or even formalisms (see https://github.com/w3c/pub-manifest/issues/98#issuecomment-539502292), we can do that more easily... And with the predominance of infra, this would probably be a better place for it.

iherman:
I must admit that, looking through the text, I am reinforced in my opinion that something _like_ the WebIDL is necessary. For my taste the move towards infra did not make the spec more easy to read (maybe even on the contrary, but that may be personal), and such an extra aid is really a good thing to have.

index.html
iherman: Actually, I just got a note that it _is_ permitted to refer to infra normatively. I guess we can leave this as is, then.

mattgarrish:
> For my taste the move towards infra did not make the spec more easy to read

No, that's what I've been struggling with the last couple of days... ;)

But by the end I did find it to be more precise than the rather loose references to javascript and generic object/arrays that we were using. Plus there is the benefit that if devs are already reading this notation in a [growing list of other specs](https://github.com/whatwg/infra/wiki/Dependent-specs), we're not expecting them to figure out something new.

(I couldn't make sense of the final data without a cleaner representation, either.)

rdeltour:
As an attempt to define the data structure using Infra, here's what it could look like:

---

A _localizable string_ is a [struct](https://infra.spec.whatwg.org/#struct) with the following items:
- value_, which is a [Javascript string](https://infra.spec.whatwg.org/#javascript-string)
- _language_, which is a [Javascript string](https://infra.spec.whatwg.org/#javascript-string)
- _direction_, which is \``ltr`\` or \``rtl`\`

(…)

A _publication manifest_ is a [struct](https://infra.spec.whatwg.org/#struct) with the following items:
- _type_, which is a [list](https://infra.spec.whatwg.org/#list) of [Javascript strings](https://infra.spec.whatwg.org/#javascript-string)
- _conformsTo_, which is a a [list](https://infra.spec.whatwg.org/#list) of [Javascript strings](https://infra.spec.whatwg.org/#javascript-string)
- _id_, which is a [Javascript string](https://infra.spec.whatwg.org/#javascript-string)
- _abridged_, which is a [boolean](https://infra.spec.whatwg.org/#boolean)
- …

---

It may not be as concise as a WebIDL dictionary, but is reasonably easy to read?

rdeltour:
Note that in my Infra definition above the types are defined as _structs_ rather than _maps_ as currently used in the PR, which seem to be a better fit…

mattgarrish:
> the types are defined as _structs_ rather than _maps_ as currently used in the PR, which seem to be a better fit…

For expressing the structure, yes, but the [conversion of JS to infra](https://infra.spec.whatwg.org/#convert-a-json-derived-javascript-value-to-an-infra-value) changes objects to maps, so the PR is correct.

rdeltour:
> the [conversion of JS to infra](https://infra.spec.whatwg.org/#convert-a-json-derived-javascript-value-to-an-infra-value) changes objects to maps

Ah right, good point. I knew I had seen that somewhere.

rdeltour:
> the [conversion of JS to infra](https://infra.spec.whatwg.org/#convert-a-json-derived-javascript-value-to-an-infra-value) changes objects to maps

However, maps can be used as the starting point and converted to structs by the processing algorithm… Anyways, it's probably bike shedding at this proof-of-concept stage 😊

mattgarrish:
> However, maps can be used as the starting point and converted to structs by the processing algorithm…

Could you do this easily? Don't structs come with a restricted set of members? The incoming data may not match the struct because of the extensibility of the manifest, so would it require a new struct that extends the base struct?

(WebIDL doesn't really capture this extensibility, either, I suppose, but if we're actually converting...)

> Anyways, it's probably bike shedding at this proof-of-concept stage

Agreed.

index.html
mattgarrish: No, this wouldn't happen because neither `lang` nor `dir` gets set again once they get a value.

Working backwards, `lang` gets set to "ar", so is no longer an empty string, and `dir` is and empty string after inspecting the first object. Then `dir` gets set to "ltr" from the next object and neither is an empty string any longer so the process ends.

index.html
mattgarrish: Yes, I don't think we ever require a single entity, but I suppose the prose fix to leave in place just in case we ever do is:

> If term expects a single entity or array of entities

index.html
mattgarrish: ya, must have had a brain fart when I originally wrote that...

index.html
mattgarrish: yes, when I ask for input it's because I generally hate what I came up with... :)

What if to combine with @rdeltour's earlier response we call the appendix "Internal Representation Notations"?

That way we could present one or more notations for the structure, if we were so inclined. And we could also make clear that the appendix is just one way of visualizing the data using a known syntax, instead of giving the impression it is normative or the only way.

mattgarrish: I ended up with "Internal Representation Data Models". Might be odd to talk about representations and then only have one, though. Will have to see how this evolves.

index.html
iherman: Yep, you are right!

index.html
iherman: I can live with that; as far as I am concerned, it is a go.

Another bike shedding issue: In some sense, with the current setup, the WebIDL stuff is not more important than the TOC algorithm (or maybe even less), so if the TOC is an appendix not a bona fide section, then the WebIDL should be, too. But I leave that to you.

index.html
iherman: This is not yet done in the latest incarnation...

iherman: (I mean in point 8.1.5)

iherman:
I am not sure what is happening, and I cannot locate it on the diff view above: It seems that part of 7.1 (setting the `name` to the value of `<title>` as a default) is missing toward the end. This is what I see on my screen (I checked it with two different browsers):

Screen Shot 2019-10-10 at 06 23 07

mattgarrish:
> It seems that part of 7.1 (setting the `name` to the value of `<title>` as a default) is missing toward the end.

No, I put a `pre` where there should be a `code`. It's saying to set name to a list with title as the only item:

> Set `processed["name"]` to `« title »`.

index.html
mattgarrish: Oops, ya, I fixed the second instance where LinkedResource was tagged, but didn't get the reference in parens. Will patch in a minute with some other tweaking of the references to linkedresource in the doc.

index.html
mattgarrish: > But I leave that to you.

You could have just said you forgot to make it an appendix when you moved it doofus... because, well, that's what I did. :)

index.html
```diff
@@ -3020,36 +2919,34 @@ <h2>Processing a Manifest</h2>
 						manifest</a>, when available.</li>
```
rdeltour: (commenting here has this specific text wasn't affected by the PR):

> _base_: a URL string that represents the base URL for the manifest.

should link back to our definition of the manifest base URL.

index.html
```diff
@@ -3476,110 +3497,147 @@ <h2>Processing a Manifest</h2>
...
 						</li>
 
 						<li id="processing-checks-pub-date">
-							<p>(<a href="#publication-date"></a>) If <var>processed["datePublished"]</var> is set and
-								its value is not a valid date or date-time per&#160;[[!iso8601]], this is a
-									<a>validation error</a>.</p>
+							<p>(<a href="#publication-date"></a>) If <var>processed["datePublished"]</var> is set and is
+								not a valid date or date-time per&#160;[[!iso8601]], <a>validation error</a>, <a
+									href="https://infra.spec.whatwg.org/#map-remove">remove</a>
+								<var>processed["datePublished"]</var>.</p>
```
rdeltour: _Note_: this comment shouldn't block this PR, it can be further discussed in #104 and —if it results in any change— dealt-with in another PR.

do we want to allow smart UA to recover from these? Some might be able to fix an invalid date string, or convert a date in natural language to the ISO format.

One possibility could be to define a generic step like "convert input to the expected type" as a UA-defined step, and either return that or remove.

I'm not sure; what do you think?

iherman: Specifically for dates this may make sense (I am a little bit afraid we create a precedence, but maybe we are fine), but the step you propose can be read as if the UA _MUST_ try to recover. This may be a step too far (let alone the fact that we would have to test this, and I am not sure how many different strings there are that can be interpreted correctly). 

"... is set and is not a valid date or date-time per [[!iso8601]], validation error. If the User Agent is able to interpret the value as a valid date or date-time, replace the value with that generated value; otherwise remove it."

Or something like that, converted to infra-speak...

mattgarrish: Doesn't this lead to interop problems if we can't define how to fix the invalid input? There are a whole host of reasons why the input could be invalid, from the data type being wrong to the date string not conforming.

iherman: Yeah... what is worse, it could lead to wrong interpretations. E.g., what date is meant by "01/02/03"? 1st of February 2003 (in the UK), 2nd of January 2001 (in the US) and, I believe, 3rd of February in 2001 in Japan (or in Hungary, b.t.w.).

Maybe we should keep it as is after all.

index.html
```diff
@@ -3361,61 +3355,85 @@ <h2>Processing a Manifest</h2>
...
+								<li>
+									<p>otherwise, if <var>normalized</var> is a <a
+											href="https://infra.spec.whatwg.org/#list">list</a>, perform this step on
+										each <a href="https://infra.spec.whatwg.org/#list-item">item</a> of the
+										list.</p>
+								</li>
+								<li>
+									<p>otherwise, if <var>normalized</var> is a <a
+											href="https://infra.spec.whatwg.org/#ordered-map">map</a>, perform this step
+										on each <a href="https://infra.spec.whatwg.org/#map-key">key</a> of the map.</p>
```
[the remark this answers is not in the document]
mattgarrish: No, looks like I pasted them into the wrong place at some point. Those two steps are supposed to come after the normalizing steps, as they're intended to recursively descend into `normalized` to run these operations on subproperties.

If we only go through the primary keys of the manifest, we'll miss normalizing properties inside LinkedResources, for example (or whatever an extension might defined).

index.html
rdeltour: > but the step you propose can be read as if the UA _MUST_ try to recover.

sorry I wasn't clear. I was suggesting a MAY at best 🙂

> Doesn't this lead to interop problems if we can't define how to fix the invalid input? 

yeah, that's why I'm not sure.It's not so uncommon to have implementation-defined behavior/processing. There's not a huge risk for interoperability I think, given the nature of the metadata. At worst there will be cases like users not getting their publications sorted in the same order on various UAs if they sort it by date, maybe this isn't too bad? Again, these are open questions, I don't have a strong opinion on this.

> it could lead to wrong interpretations

Right. I think that given a language and a date string, a lot can be done (e.g. "Dec 2019", "01/02/03" are not unambiguous in US English). I'm more worried of implementation starting using a default date instead of removing the value if they can't make sense of the authored date.

As always it would be interesting to see if "invalid" dates are commonly found in today's EPUB's for instance, to have an idea whether this would be useful at all.

In any case, keeping it as is may be safer for a first version?

index.html
```diff
@@ -3361,61 +3355,85 @@ <h2>Processing a Manifest</h2>
...
-							<p>If typeof(<var>current</var>) is <code>Object</code>&#160;[[!ecmascript]], perform this
-								step on each property of the object.</p>
+						<li id="processing-normalize-urls">
+							<p>(<a href="#value-url"></a>) If <var>term</var> expects a <a href="#value-url">URL</a></p>
+							<ol>
+								<li>
+									<p>if <var>normalized</var> is a <a href="https://infra.spec.whatwg.org/#string"
+											>string</a> that is not an <a data-cite="!url/#absolute-url-string">absolute
+											URL string</a>, resolve it to an absolute URL string using the value of
+											<var>base</var>&#160;[[!rfc1808]].</p>
```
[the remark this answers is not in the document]
mattgarrish: Ya, I can't track down where 1808 came from, but since we use URL standard everywhere else, I don't see why we wouldn't use it for resolving.

index.html
iherman: > In any case, keeping it as is may be safer for a first version?

Yes

mattgarrish:
I just pushed a change to unlink the webidl from the property definitions, but could be convinced to undo this.

Given that we've made the idl informative and moved to an appendix, the bidirectional linking is confusing.

mattgarrish:
In addition to @rdeltour's comments, I also added the default profile URL from #101 (I still don't quite understand how this will work, but we can tweak that).

Also, per https://github.com/w3c/audiobooks/issues/27, I removed `alternate` from the step to expand LinkedResources and put it into the validation steps to make sure that `url` and `encodingFormat` are always set.

The only way we could allow alternate resources to be strings is if get into inferring media types from extensions. It may be more complicated than that, though, as we'd have to lower the requirement for `encodingFormat` to a should (i.e., we'd have to bless media type guessing).

We could permit guessing to avoid alternate resources being thrown out, but like the other issue with dates it gets into how much work we want to push onto user agents.

mattgarrish:
Hm, I should have double-checked the definition. I was thinking encodingFormat was required except for HTML, but now I see it's just:

> Non-HTML resources SHOULD be expressed as LinkedResource objects with their encodingFormat values set.

So guess we can do both. Expand strings and also validate that encodingFormat is set (but not omit if not).

iherman:
> I just pushed a change to unlink the webidl from the property definitions, but could be convinced to undo this.
>
> Given that we've made the idl informative and moved to an appendix, the bidirectional linking is confusing.

Indeed. However, there is a strange effect now (respec is getting in our way): in, say, A.1.1., the terms themselves (like `abridged`) are active links which link to, I believe, themselves, ie, not back into the text. Which is a bit strange.

I wonder whether we should not get rid of respec's WebIDL processing here...

Also: you have a subsection A.1 for WebIDL. Does it mean that we would want to put other formalisms as "A.2"?

iherman:
> Also, per w3c/audiobooks#27, I removed alternate from the step to expand LinkedResources and put it into the validation steps to make sure that url and encodingFormat are always set.

Is this related to #105 ? My feeling is that there should be no exception for audiobooks, in fact...

mattgarrish:
> I wonder whether we should not get rid of respec's WebIDL processing here...

That works for me.

> Does it mean that we would want to put other formalisms as "A.2"?

Leaving open that possibility, yes. We can always revise at the end of CR if we still only have webidl to show.

mattgarrish:
> My feeling is that there should be no exception for audiobooks, in fact...

Ya, I screwed that up by not double-checking what I thought I knew. I put back the expansion of strings, but I'm still confused about why HTML resources get a pass from `encodingType`. As I just mentioned in #105, either warnings for all or no warnings.

iherman:
> Leaving open that possibility, yes. We can always revise at the end of CR if we still only have webidl to show.

As a new convert to Typescript I believe it would be useful:-) but I am not sure it would bring anything new, so it may be unnecessary.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-10-14-pwg#section3).

- No actions or resolutions
update manifest processing to use infra types<br/>
Garth Conboy: See Issue #101<br/>
Garth Conboy: See Issue #98<br/>
Ivan Herman: See PR #103<br/>
Garth Conboy: There are 2 issues in this pull request - this is also one that Ivan and Matt have…<br/>
… Does anyone on the call want to run through this and hope to get to a consensus?<br/>
Ivan Herman: The work was done by Matt.<br/>
… I would prefer Romain to comment on this as he was the one initiating this whole work and can give a more appropriate background.<br/>
Romain Deltour: Basically it originated from trying to get rid of the webIGL as a way to describe a datastructure. In trying to harmonize the various manifests in the W3C, we wanted to use infra-types to represent the data structure. Matt’s editing the spec to replace webIGL for infra types.<br/>
Ivan Herman: See Infra standard<br/>
Romain Deltour: The WebIDL representative is moved to informative edits. A few different issues were raised when we looked at the algorithm. I think it’s clearer now, but we need more reviews there. It will be interesting to have more.<br/>
Garth Conboy: Looking at the PR, Romain you have one requested change outstanding and Ivan has approved it.<br/>
Ivan Herman: I am now repeating what Dave said earlier on the other PR. This was a pretty large change that made the document better and Matt did great work. Now it becomes difficult to discuss other issues and details with this as a PR.<br/>
Romain Deltour: I think my requests have been addressed in Matt’s latest changes<br/>
Ivan Herman: I would be in favor of merging this ASAP (as soon as Matt is comfortable) then we can comment and modify once it’s in. Again, this is something he should do…<br/>
Romain Deltour: +1 on merging and doing more reviews/changes based on Matt’s work<br/>
Garth Conboy: Sounds good for me. If there are subsequent changes, they will be smaller. Barring objection, why don’t we go ahead and put that as a request for Matt…<br/>
</details>

Issue #109, opened 2019-10-13
value type check complicated and draconian


mattgarrish:
I'm wondering how useful the validity check for expected value types is.

If you take the test literally, if an author array has one invalid entity in it, for example, then you have to throw out the entire property.

We have value checks following this one that will raise the most glaring errors, and only remove the offending items when an array/object is expected.

I'm thinking it might be better just to look and see what properties we don't have any tests for (a11y and maybe a couple of others) and see how to test their values in a more atomic way.

iherman:
> If you take the test literally, if an author array has one invalid entity in it, for example, then you have to throw out the entire property.
>
> We have value checks following this one that will raise the most glaring errors, and only remove the offending items when an array/object is expected.

I agree it is a bit confusing, although, indeed, either you have to remove the entire property or be clear that, at the end of the checking/cleaning phase an array may become empty. Maybe as a last step of the checking sequence we can clearly say that, and, if we get an empty array for some entries (probably the reading order) then this is actually a fatal error.

---

However, it is a bit difficult to discuss this now because the current editors' draft and the relevant parts of #103 have become fairly different. I would think we should get an approval from the WG to merge #103 and, if that happens, we have a cleaner base to discuss all these.

mattgarrish:
> Maybe as a last step of the checking sequence we can clearly say that, and, if we get an empty array for some entries (probably the reading order) then this is actually a fatal error.

Yes, this is another problem related to this that I was going to bring up. This happens right now anyway, as invalid URLs are removed after this step so the reading order can end up empty without an error being raised.

The same happens for entities, as all the names could get removed by a later step leaving an empty array.

There's also a lot of redundancy, as we later check if abridged is a boolean, but that's what this value type step is supposed to check. If dates and length are invalid, that's also caught later, so ensuring they are strings in this step also doesn't do a lot (other than maybe preventing arrays?).

A generic step like this would be nice, but I think it's better that we define all the validation/processing steps.

But, yes, I'm opening these as separate issues so hopefully we can merge what we have in #103 and deal with these separately. Just problems I'm seeing as I try to implement the algorithm.

mattgarrish:
At the very least, the step needs to be broken out so objects and arrays are dealt with separately:

> (§ 2.3 Value Categories) Check properties with known value types as follows:
> 
> - if `term` expects an array:
>   - if `termValue` is not an array, validation error, remove `processed[term]`.
[...]

PR #111, opened 2019-10-13
experimental processor

mattgarrish:
No claims of elegance or comprehensiveness of validity checks, but here's the code I've been tinkering with to test the processing algorithm. Includes the clarification I made in https://github.com/w3c/pub-manifest/issues/109#issuecomment-541412866 but doesn't address the redundancies and missed checks that still need to be looked at.

Needs to be run from localhost or off a server. You can maybe run locally if you put the manifest in the script tag, but I have a feeling the URLs will all be invalid since we require http(s).


mattgarrish:
Here's an example of the output: https://raw.githack.com/w3c/pub-manifest/experiment/internal-rep-processor/experiments/manifest-to-internal-processor/test.html

I might pretty it up later if my brain ever recongeals from this.

Issue #115, opened 2019-10-15
single localizable string?


mattgarrish:
I noticed for `accessibilitySummary` and `LinkedResource/description` that the expected value category is "LocalizableString" and not "Array of LocalizableStrings".

I have to believe this is a mistake, as why couldn't you provide these in more than one language?

I can't envision a case of a property with a localizable string that wouldn't take an array of them.

iherman:
I would defer the `accessibilitySummary` to @avneeshsingh and/or @clapierre. I agree on the `description` piece.

avneeshsingh:
It makes sense to have multiple strings in different languages. We also need to ensure that it is in sync with schema.org definition.
When we worked with schema.org people in 2016, AccessModeSufficient was the only accessibility metadata that was identified for array representation. 
So, just a caution, we are near CR, so the changes that we make should be in sync with schema.org.

iherman:
Thanks @avneeshsingh, I was wondering about the same thing. B.t.w., `description` is also defined similarly.

However... I believe all properties can use several text values out of the box with schema.org, as long as the order is irrelevant. The [json-ld context file for schema.org](https://schema.org/docs/jsonldcontext.json) does not preclude that, and the [google structure date testing tool](https://search.google.com/structured-data/testing-tool/u/0/) does not complain either. I tried the following test:

```
<script type='application/ld+json'>
{
"@context":"http://schema.org",
"type":"CreativeWork",
"name": "somebody",
"description": ["one","two"],
"accessibilitySummary": ["this", "and that"]
}
</script>
```

and it worked without further ado. Ie, we may be o.k.

mattgarrish:
Right, this appears to be a case of our being more limiting than schema.org.

We've just thought of the summary in terms of a single language because most publications are not multilingual, but even in epub there's nothing to prevent you from adding it multiple times to the package document for different languages:

```
<meta property="schema:accessibilitySummary" xml:lang="en">...</meta>
<meta property="schema:accessibilitySummary" xml:lang="fr">...</meta>
```

mattgarrish:
The summary is the only field this applies to out of the accessibility metadata, just to be clear. The rest all expect predefined tokens (literals).

AccessModeSufficient is another unique case in that each individual entry is a list, but that's a different case than being able to represent strings in multiple languages.

iherman:
[...]

Issue #113, opened 2019-10-15
mixing value type/category


mattgarrish:
I've noticed we interchange value type/category throughout the specification, even in the definition tables.

It looks like "value category" is the preferred term, since that's what we've named the section that describes them, but I need to go through and harmonize the language.

mattgarrish:
Closing this with #116

Issue #110, opened 2019-10-13
accessmodesufficient value type confusing


mattgarrish:
The accessModeSufficient property expects one or more ItemList, but the value category says "Array of Literals".

We don't have a value type for an ItemList, so I'm not sure how to correct this. It's sort of like we want an "Array of Array of Literals".

iherman:
Hm. This is now clearer in https://schema.org/accessModeSufficient (I remember this wasn't very clear before). We should probably refer to the Schema.org page, if we can, instead of reproducing the definition ourselves...

mattgarrish:
It's not the definition that is the problem, though, but the value type category.

One thought would be to put "(None)" into the field and bypass trying to check this one with the value type step. The problem is it wouldn't normalize the value to an array, which seems necessary.

Or maybe we should define a generic 'Object' type and call it an "Array of Objects" so all that gets check is that each entry is a map. A later step could then verify the maps all represent ItemLIsts.

iherman:
> Maybe we should define a generic 'Object' type and call it an "Array of Objects" so all that gets checked is that each entry is a map. A later step could then verify the maps all represent ItemLIsts.

Actually, this may be a good idea for possible future extensions.

mattgarrish:
Closing this with #116

Issue #108, opened 2019-10-12
split processing functions


mattgarrish:
Instead of saying to call "this step", it might make it easier to read to define the relevant steps as their own functions. It might be more obvious when properties are being recursively descended into, too.

iherman:
I am not sure I understand...

mattgarrish:
Sorry, I'm trying to write some code to match the algorithm to check that it works as expected, so making issues that are probably unreadable outside my brain.

An example here would be all the normalizing steps. Instead of having all these steps in the main algorithm, we could split them out. So section 6 might just say.

> For each `term` → `termValue` of `manifest`:
> 1. Let `normalized` be the result of calling normalize data given `term` and `termValue`.
> 2. Set `processed[term]` to `normalized`.

After the main processing algorithm we define "normalize data" as the series of steps to run and we can use generic variables.

That way when we need to descend into objects to process additional values, we say to run "normalize data" using the keys/values of that object instead of to run "this step" again. The function becomes self-contained and clearer, so we don't have to make sense of references to `manifest`/`processed` when we're really somewhere below that inside an object.

iherman:
Ah... It is difficult to see in abstract whether it is helpful or not; my gut feeling says it is. But yield completely to your assessment on this!

mattgarrish:
Closing this with #116

PR #116, opened 2019-10-16
Updates to processing algorithm and other fixes

### LINKED ISSUE #101: Lack of conformsTo too draconian?

mattgarrish:
This PR makes the following changes:

- updates the processing algorithm to split it into more atomic functions where appropriate (makes the main algorithm clearer and also makes it easier for profiles to extend specific functions instead of step numbers).
- updates description and accessibilitySummary to arrays of localizable strings.
- standardizes the name "value category".
- adds a generic object category and updates accessModeSufficient to reference it.
- adds prose for reading systems to inspect reading order for unknown profiles and react accordingly.

It should fix #101 #106 #108 #109 #110 #113 and #115

I'm going to update the experimental processor code next just to make sure I didn't break anything splitting up the algorithm.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/116.html" title="Last updated on Oct 18, 2019, 1:59 PM UTC (b768200)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/116/07f264d...b768200.html" title="Last updated on Oct 18, 2019, 1:59 PM UTC (b768200)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -858,7 +858,7 @@
 							href="https://url.spec.whatwg.org/#absolute-url-string">absolute-URL strings</a> using a <a
 							href="https://url.spec.whatwg.org/#concept-base-url">base URL</a>&#160;[[!url]].</p>
 
-					<p id="manifest-base-url">The base URL for relative-URL strings is determined as follows:</p>
+					<p>The <dfn>base URL</dfn> for relative-URL strings is determined as follows:</p>
 
 					<ul>
 						<li>In the case of an <a href="#manifest-embed">embedded manifest</a>, it is the <a
@@ -2839,8 +2839,7 @@
 
 				<ul>
 					<li><var>text</var>: a UTF-8 string representing the manifest;</li>
-					<li><var>base</var>: a URL string that represents the <a href="#manifest-base-url">base URL for the
-							manifest</a>.</li>
+					<li><var>base</var>: a URL string that represents the <a>base URL</a> for the manifest.</li>
 					<li><var>document</var>: the <a data-cite="html#the-document-object">HTML Document (DOM)
 						Node</a>&#160;[[!html]] of the document that <a href="#manifest-discovery">references the
 							manifest</a>, when available.</li>
@@ -2857,12 +2856,6 @@
 						manifest</dfn>, run the following steps:</p>
 
 				<ol>
-					<li id="processing-init">
-						<p>Let <var>processed</var> be an empty <a href="https://infra.spec.whatwg.org/#ordered-map"
-								>map</a> that will contain the <a href="#dfn-internal-representation">internal
-								representation</a> of the manifest.</p>
-					</li>
-
 					<li id="processing-json-infra">
 						<p>Let <var>manifest</var> be the result of parsing <a
 								href="https://infra.spec.whatwg.org/#parse-json-into-infra-values">JSON into Infra
@@ -2948,39 +2941,31 @@
 						</details>
 					</li>
 
+					<li id="processing-create-processed">
+						<p>Let <var>processed</var> be an empty <a href="https://infra.spec.whatwg.org/#ordered-map"
+								>map</a> that will contain the <a href="#dfn-internal-representation">internal
+								representation</a> of the manifest.</p>
+					</li>
+
 					<li id="processing-normalize">
 						<p><a href="https://infra.spec.whatwg.org/#map-iterate">For each</a>
 							<var>term</var> → <var>value</var> of <var>manifest</var>, set <var>processed[term]</var> to
 							the result, when successful, of calling <a>normalize data</a> given <var>term</var>,
-								<var>value</var>, <var>lang</var> and <var>dir</var>. If failure is returned, do not add
-								<var>term</var> to <var>processed</var>.</p>
-					</li>
-
-					<li id="processing-defaults">
-						<p>Set <var>processed</var> to the result of running <a>add HTML defaults</a>, when successful,
-							given <var>processed</var> and <var>document</var>. If failure is returned, terminate
-							processing, return failure.</p>
+								<var>value</var>, <var>lang</var>, <var>dir</var> and <var>base</var>. If failure is
+							returned, do not add <var>term</var> to <var>processed</var>.</p>
 					</li>
 
 					<li id="processing-checks">
 						<p>Set <var>processed</var> to the result of running <a>data validation</a> given
 								<var>processed</var>.</p>
 					</li>
-
-					<li id="processing-remove-empty-arrays">
-						<p><a href="https://infra.spec.whatwg.org/#map-iterate">For each</a>
-							<var>term</var> → <var>value</var> of <var>processed</var>, if running <a>remove empty
-								arrays</a> given the variables <var>term</var> and <var>value</var> returns failure, <a
-								href="https://infra.spec.whatwg.org/#map-remove">remove</a>
-							<var>processed["term"]</var>.</p>
-						<details>
-							<summary>Explanation</summary>
-							<p>As the processing of the manifest involves removing invalid values at various stages, the
-								final data structure might end up with some lists that not no longer contain any values.
-								This step iterates back over the data and removes any such empty lists.</p>
-						</details>
+					
+					<li id="processing-defaults">
+						<p>Set <var>processed</var> to the result of running <a>add HTML defaults</a>, when successful,
+							given <var>processed</var> and <var>document</var>. Otherwise, terminate processing, return
+							failure.</p>
 					</li>
-
+					
 					<li id="processing-end">
 						<p>Return <var>processed</var>.</p>
 					</li>
@@ -2996,9 +2981,11 @@
 						the direct property of the Publication Manifest, but a single URL literal when used in other
 						objects.</p>
 
-					<p>To <dfn>normalize data</dfn> for a property <var>term</var>'s <var>value</var>, with optional <a
-							href="https://infra.spec.whatwg.org/#string">strings</a>
-						<var>language</var> and <var>direction</var>, run these steps:</p>
+					<p>To <dfn>normalize data</dfn> for a property <var>term</var>'s <var>value</var>, with the <a
+							href="#manifest-lang-dir-global">global language</a>
+						<var>lang</var>, <a href="#manifest-lang-dir-global">global direction</a>
+						<var>dir</var> and <a>base URL</a>
+						<var>base</var>, run these steps:</p>
 
 					<ol>
 						<li id="normalize-init">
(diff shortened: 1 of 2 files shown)


mattgarrish:
I also split the processing section in three, as it was getting bloated. There's now introduction and error handling subsections.

iherman:
To repeat what I said in https://github.com/w3c/pub-manifest/issues/101#issuecomment-542123710: the algorithm yield the _profile_ value, that is then part of the generated data structure. As a consequence, the WebIDL should have a `profile` member, and no `conformsTo` one...

index.html
```diff
@@ -3254,38 +3316,46 @@ <h2>Processing a Manifest</h2>
 								result of running <a href="#processing-normalize">this step</a> on
 								<var>keyValue</var>.</p>
 						</li>
```
iherman: Just to be precise: running this step on keyValue, lang, and dir. (ie, the input lang and dir as forwarded to the recursive call)

iherman:
I wonder whether it is possible to add explanation blocks to the "functions", mainly the last ones. By the time the reader gets there, he/she may be a bit lost as for the goals of these and the fit the full picture.

mattgarrish:
>  the algorithm yield the _profile_ value, that is then part of the generated data structure. As a consequence, the WebIDL should have a `profile` member, and no `conformsTo` one...

Adding another member is easy enough, but we shouldn't get rid of `conformsTo` since it's not exclusively about setting a profile. It's also a means of identifying conformance with wcag or other possible standards, so that information shouldn't still be retained.

mattgarrish:
I've just made a few more changes to fix up the algorithm. The only prominent one is to move the harvesting of html defaults to the last step. Until the data is processed and validated (and invalid data removed), there's no way to know if `name` and `readingOrder` could benefit from this step (i.e., it can be useful both if values are intentionally omitted or are invalid).

index.html
mattgarrish: Yes, and base. This is corrected in the latest commit.

index.html
```diff
@@ -3361,170 +3432,301 @@ <h2>Processing a Manifest</h2>
...
+												<var>item</var> from <var>value</var>.</p>
+										</li>
+										<li>
+											<p>if <var>item</var> is a <a
+													href="https://infra.spec.whatwg.org/#ordered-map">map</a>, <a
+													href="https://infra.spec.whatwg.org/#map-iterate">for each</a>
+												<var>key</var> → <var>keyValue</var> of <var>item</var>, if
+													<var>key</var> has an expected type, set <var>key</var> to the
+												result of running <a href="#category-check">check category type</a>
+												given the variables <var>key</var> and <var>keyValue</var>. If the
```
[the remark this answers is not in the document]
mattgarrish: Yes, fixed now.

iherman:
> Adding another member is easy enough, but we shouldn't get rid of conformsTo since it's not exclusively about setting a profile. It's also a means of identifying conformance with wcag or other possible standards, so that information shouldn't still be retained.

Oops, I forgot about this. You are right

mattgarrish:
> I wonder whether it is possible to add explanation blocks to the "functions"

I'm going to do this now that it seems stable, but I also just made a commit that groups the functions together by how they're used. The first level are the ones called by the main processing algorithm, and then beneath each of those are the definitions for the specific functions they call.

iherman:
The context specific nature of some of the terms (like `url`) is not reflected in the "normalize data" function. Are the values of `term` and `value` enough as input arguments for that function? Shouldn't we also forward a value of a "context", ie, the value of `manifest` when calling from the "processing a manifest" and the value of `normalized` when recursively invoked from (8)?

iherman:
The normalization step do not seem to take care of the possibility of having, as an expected value, a single Localizable String (and turn a simple value into one). Or do we want to remove that category (this is related to #115)?

mattgarrish:
> The normalization step do not seem to take care of the possibility of having, as an expected value, a single Localizable String

Yes, that's what I was asking in #115. When would you restrict a localizable string to only one locale at the specification level? I haven't been able to think of a use case, as it seems to impact i18n if you do. If you have a localizable string, you seem to require an array.

We do convert single strings into an array with only the one localized string, but never into a term with just a localizablestring object as its value.

We could perhaps make clear in the localizable string definition that whenever you use one you have to allow an array.

dauwhe:
> If manifest is not a map, fatal error, return failure.

Is it possible for a syntactically valid JSON file to not be a map?

mattgarrish:
> Is it possible for a syntactically valid JSON file to not be a map?

Not that I'm aware of. A JSON object is a set of key/value pairs which is what a map represents.

I checked the translation to infra types with the WebDriver spec to try and match up the conversion prose.

mattgarrish:
> otherwise what system do we use to differentiate?

Do we just keep sending through the initial key into each recursive call, so if you find `url` anywhere inside of `readingOrder` it has to be a single value?

mattgarrish:
>  I worry that this is invalid:

Yes, it is, but the algorithm is only checking what the spec requires: https://w3c.github.io/pub-manifest/#manifest-context

I believe the order is necessary because our context inherits from schema.org, but @iherman can speak better to that. (But a new issue would really help me track what's left to do here :)

iherman:
>> If manifest is not a map, fatal error, return failure.
> Is it possible for a syntactically valid JSON file to not be a map?

Yes. It can be an array.

iherman:
>> If, depending on context, term expects an array ...
>
> But, related to the other question about single localizable strings, when would you not allow more than one entity, or disallow more than one linked resource? We don't currently have any use cases for these, so do we make the algorithm account for scenarios that may never play out?

The only place it came up, afaik, is with `url`. It does not make any sense to have multiple `url`-s for a linked resource, but it seems to be o.k. to have several `url`-s for a publication.

The only (very weak!) use case I have for the latter is the duality of short name based URI-s for a W3C TR document and its full blown, dated one. Although I would probably encode it using `id` for the short named one...

iherman:
> Although, trying to think this through, what meaningful context can you send to the function? What distinguishes url coming in in the initial call from url being encountered recursively? Maybe if a type is set you'll have some context for the nested case, but otherwise what system do we use to differentiate?

In my (coming, but slowly...) Typescript implementation the only information you need is its type (in the Typescript sense). I am not sure how to put that information into infra in general...

iherman:
> I believe the order is necessary because our context inherits from schema.org, but @iherman can speak better to that. (But a new issue would really help me track what's left to do here :)

I think the order is relevant; there are some schema.org terms for which we define some additional information in the context. Typical case is to make it clear that the order of the authors is significant (in JSON-LD parlance we add a "@container:@list" to the reference). If the schema.org context came second, it would wipe this out.

mattgarrish:
> The only place it came up, afaik, is with `url`. It does not make any sense to have multiple `url`-s for a linked resource, but it seems to be o.k. to have several `url`-s for a publication.

Right, that's why I'm wondering if we should just put in the array step that whether to convert a value is context-dependent but leave the rest of the algorithm alone (i.e., for linkedresource, entity and localizablestring, keep the expectation that they will be arrays).

There are probably other cases where single/array may factor into processing, but it seems like something we can leave for extension steps to cover.

iherman:
You mean we make an exception for href or for the array? But either way, the information whether that step should be applied is context dependent...

I guess I do not really understand what you mean:-(

mattgarrish:
> You mean we make an exception for href or for the array?

We make an exception for whether `url` gets converted to an array. If the context is `url` (the top-level instance), it allows an array. If the context is `readingOrder`, `resources`, `links` or `alternate`, then there is no conversion to an array (we'd have to similarly check this at the value category step, too).

But after the step to convert the incoming data to an array, we have three steps that are array-dependent (for an "array of linkedresources", etc.). I say we leave these alone and not make them, for example, "a linkedresource or array of linkedresources" and have to handle both cases in the algorithm. We don't have any cases where a property accepts single values of these.

mattgarrish:
Another sad realization I'm having while looking at adding explanations is that it no longer makes a lot of sense to use javascript examples when we're talking about infra types. They should probably be swapped to the new syntax... :'(

mattgarrish:
The last commit has a couple of major changes:

- it adds context variables to function calls and also adds a definition at the start of the algorithm so the purpose of the context doesn't repeatedly have to be explained.
- it adds a `profile` term to internal rep;

I don't like reusing "context", but can't think of a better name. Having the definition, and linking to it from the function calls, helps a bit, though.

And with the change to note that LocalizableString always becomes an array, I think I'm now up to date with all the comments in this thread. Let me know if there are still issues with the changes I've made, though

iherman:
I would say merge this, it is a great equilibrium point.

As I go ahead with my typescript toy implementation I may hit issues, but we can treat them as they come.

B.t.w.:

> Another sad realization I'm having while looking at adding explanations is that it no longer makes a lot of sense to use javascript examples when we're talking about infra types. They should probably be swapped to the new syntax... :'(

I see you have made the changes and I am not sure it was necessary... I still find the JS/JSON syntax _way_ more readable than infra so, as a reader, the examples may be less helpful to me now than they were. But maybe this is because I am an old-skool person who has not yet grasped the beauties of infra...

iherman:
Minor: 4.3.1./3 (setting entities) if the value is an Array, it should also be a validation error. Actually, this may be the case for (4) and (5), too

iherman:
> Actually, this may be the case for (4) and (5), too

Sorry, it is there for (4) and (5) (the terminology is 'list' not 'array' :-)

mattgarrish:
> But maybe this is because I am an old-skool person who has not yet grasped the beauties of infra...

The use of guillemets with both lists and maps makes it more confusing than it probably needs to be, but you get used to it.

Try writing it without a french keyboard or an arrow character, though... :(

mattgarrish:
> (the terminology is 'list' not 'array' :-)

Normally I'd agree with you, but not in the normalization function. We're referring to our value categories when talking about what a term expects, so it's clearer to use "array" than link "list" to our non-infra definition.

But, yes, a list in a list of entities should be an error. I'll add that.

iherman:
Point 8 in the normalization function is, I believe, wrong. It says:

> if normalized is a list, for each item of normalized that is a map, for each key → keyValue of normalized, set key to the result of running normalize data given key, keyValue, lang, dir, base and context.

I believe it should say

> if normalized is a list, for each item of normalized that is a map, for each key → keyValue of item, set key to the result of running normalize data given key, keyValue, lang, dir, base and item.


- the key/keyValue should refer to the item (in the list) and not to normalized
- the context set for the recursive normalization should be item (ie the map in the list), and not the context that was used when calling normalization

Also the return value of the normalization may be a 'failure'. Just as for the top level, the following remark should also be added somehow. That means it should be removed from the list.

(When the value is a map, that means that 'normalized' will get the failure value, which will return failure to the upper level caller. Although it may be good to spell this out.)

mattgarrish:
> * the key/keyValue should refer to the item (in the list) and not to normalized

Yes, that's wrong.

> * the context set for the recursive normalization should be item (ie the map in the list), and not the context that was used when calling normalization

The context should be `key`, I believe, not the whole object, as that way you always have the owner term to compare against (e.g., `readingOrder` would be the context for all the linkedresources in its array, and then `alternate` would become the context when you begin processing all the linkedresources within that term).

mattgarrish:
I've made the latest batch of fixes in the last commit and also added a check to remove `@context` during the normalization step.

I'm going to merge this now, as it's grown beyond its useful life as a PR now. Time to get back to more atomic commits.

PR #118, opened 2019-10-17
Fix required schema properties

mattgarrish:
This got forgotten, but it updates the schema to reflect the actual authoring requirements - context and conformsTo.


iherman:
Merged this, it is an obvious mistake

PR #128, opened 2019-10-20
Additional algorithm updates

mattgarrish:
I found a few more issues with the algorithm while updating the test code:

- a couple of steps were referring to incorrect variable names;
- the recursion step in validation checks has to occur before any of the validity checks (except type), as the validity checks have to work from the bottom up. if not, invalid data won't yet have been removed when some tests are run.
- similarly, it's necessary to remove invalid localizable strings before checking, for example, whether an entity has a name, so I had to move it up.
- we weren't checking whether localizable strings have a `value`, so I added that with the language/direction tests
- the absolute url function wasn't checking/returning failures from url parser; I also simplified this a bit as it doesn't need recursion.
- I added some explicit "otherwise, do nothing" steps where we ending with an "otherwise, if";

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/128.html" title="Last updated on Oct 21, 2019, 6:53 PM UTC (2a2f0ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/128/bfdeae3...2a2f0ab.html" title="Last updated on Oct 21, 2019, 6:53 PM UTC (2a2f0ab)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2843,6 +2843,9 @@
 						<code>Organization</code> and <code>LinkedResource</code>. <a>Profiles</a> MAY extend this list
 					to include additional object types.</p>
 
+				<p>If a context is not provided to a function, the term being processed is considered a direct child of
+					the manifest.</p>
+
 				<p class="note">When extending the list of recognized profiles, the <a>data normalization</a> function
 					might also need to be extended to ensure that all objects have their type specified (e.g., if
 					authoring conveniences do not require an explicitly-typed object always be authored).</p>
--- a/index.html
+++ b/index.html
@@ -2846,9 +2846,9 @@
 				<p>If a context is not provided to a function, the term being processed is considered a direct child of
 					the manifest.</p>
 
-				<p class="note">When extending the list of recognized profiles, the <a>data normalization</a> function
-					might also need to be extended to ensure that all objects have their type specified (e.g., if
-					authoring conveniences do not require an explicitly-typed object always be authored).</p>
+				<p class="note">When extending the list of recognized profiles, the <a>normalize data</a> function might
+					also need to be extended to ensure that all objects have their type specified (e.g., if authoring
+					conveniences do not require an explicitly-typed object always be authored).</p>
 			</section>
 
 			<section id="processing-algorithm">
@@ -3884,8 +3884,8 @@
 									</li>
 								</ol>
 								<p class="note">This step currently only exists for use by <a>profiles</a>. The
-									properties defined in this specification all accept <a href="#value">arrays</a> of
-									objects.</p>
+									properties defined in this specification all accept <a href="#value-array"
+										>arrays</a> of objects.</p>
 							</li>
 							<li id="verify-type-invalid">
 								<p>Otherwise, if, depending on the <var>context</var>, <var>value</var> does not have
--- a/index.html
+++ b/index.html
@@ -2830,6 +2830,24 @@
 					the manner in which this is done.</p>
 			</section>
 
+			<section id="processing-contexts">
+				<h3>Processing Contexts</h3>
+
+				<p>Some steps in the <a>processing algorithm</a> depend on the expected value category of a term, so the
+					context in which a term is used can affect processing (e.g., <a href="#address"><code>url</code></a>
+					expects an <a href="#value-array">Array</a> (of URLs) only when the direct property of the
+					Publication Manifest). To differentiate these uses a <dfn>context</dfn> is provided to certain
+					function calls. This context is set to the type of object that initiates the processing call.</p>
+
+				<p>The default list of <dfn>recognized types</dfn> includes <code>Person</code>,
+						<code>Organization</code> and <code>LinkedResource</code>. <a>Profiles</a> MAY extend this list
+					to include additional object types.</p>
+
+				<p class="note">When extending the list of recognized profiles, the <a>data normalization</a> function
+					might also need to be extended to ensure that all objects have their type specified (e.g., if
+					authoring conveniences do not require an explicitly-typed object always be authored).</p>
+			</section>
+
 			<section id="processing-algorithm">
 				<h3>Generating the Internal Reperentation</h3>
 
@@ -2845,14 +2863,6 @@
 
 				<p class="note">This algorithm does not define how the manifest is discovered and obtained. The steps by
 					which to do so are defined by each <a>digital publication</a> format.</p>
-
-				<p>Some steps in this algorithm depend on the expected value category of a term, so the context in which
-					a term is used can affect processing (e.g., a <a href="#address"><code>url</code></a> expects an <a
-						href="#value-array">Array</a> (of URLs) only when the direct property of the Publication
-					Manifest). To differentiate these uses a <dfn>context</dfn> is provided to certain function calls.
-					This context is set to the type of object that initiates the processing call. <dfn>Recognized
-						types</dfn> are <code>Person</code>, <code>Organization</code> and
-					<code>LinkedResource</code>.</p>
 
 				<p>To <dfn data-lt="processing algorithm|processing the manifest|processing a manifest">process a
 						manifest</dfn>, run the following steps:</p>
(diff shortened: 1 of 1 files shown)


mattgarrish:
One additional tweak I made to the global checks function is to return the value right away if the term being processed doesn't have a known value category. Given that all the subsequent tests require knowledge of the data type, there doesn't seem to be any point going exploring into unknown terms. It could lead to false positives on term names.

index.html
```diff
@@ -3400,33 +3418,24 @@ <h4>Convert to Absolute URL</h4>
...
 						</ol>
 						<details>
 							<summary>Explanation</summary>
 							<p>This function checks if the value sent into it is a relative URL, and, if so, converts it
-								to an absolute URL. If the value is a <a href="https://infra.spec.whatwg.org/#list"
-									>list</a>, each item in the list is recursively passed to this function to inspect
-								it.</p>
-							<p>If the value passed to the function is not a string or a list, a failure is returned.
-								Otherwise, the updated value is returned.</p>
+								to an absolute URL.</p>
```
iherman: To be checked... but I believe the URL parser that we refer to here does not check whether the resulting URL is an https or http URL. On the other hand, our spec says that we restrict these URLs to those two. 

If the referred WHATWG function indeed does not check this, then our function must have an addition check on the scheme. If it does, it is worth emphasizing this in the explanation.

index.html
```diff
@@ -3550,15 +3559,92 @@ <h4>Global Data Checks</h4>
 								<p>(<a href="#properties-value-categories"></a>) If <var>term</var> has a known <a
 										href="#properties-value-categories">value category</a>, set <var>value</var> to
 									the result of calling <a>verify value category</a>, when successful, given the
-									variables <var>term</var>, <var>value</var> and <var>context</var>. Otherwise,
-									return failure.</p>
+									variables <var>term</var>, <var>value</var> and <var>context</var>. If failure is
+									returned, return failure.</p>
+								<p>Otherwise, reutrn <var>value</var>;</p>
```
iherman: reutrn -> return

index.html
```diff
@@ -3662,40 +3723,6 @@ <h4>Global Data Checks</h4>
...
-									<p>In order to ensure that all the properties in the manifest get processed, this
-										step recursively checks each entry for additional map entries to process. If the
-										value is a list, each item is inspected to determine if it is a map that can be
-										processed.</p>
-								</details>
-							</li>
-
 							<li id="global-end">
 								<p>Return <var>value</var>.</p>
 							</li>
```
iherman: The explanation at the end of the global data check is empty...

index.html
mattgarrish: Right, the url parser doesn't require a specific scheme. I'll add an extra step into the function to also check the scheme of the returned url.

index.html
mattgarrish: Oh, ya, I was initially going to try and write generic explanations for the functions, but it seemed like it would be too late to have them at the end so I put the explanation where the function is called and annotated each step instead. Forgot to take that one back out.

iherman:
An additional note:

In 4.3.2.1(2)

- in both branches we have to say what the context in when calling (recursively) global data check.
- the second branch seems to be a bit mixed up. Shouldn't it say:

> key → keyValue of item, set item[key] to the result of running global data checks given key and keyValue.

(Ie, s/value/item)

I actually wonder about the value of context in the various functions. It seems that the context value is set to a 'key', see, e.g., 4.3/(6). In my implementation what I really needed is not that key, but the type (I work in typescript) of the environment; in that call, for example, I simply forward 'processed' as a context (and, in fact, what I really need is its type). Are we sure that the current choices are the right ones?

iherman:
(Forgot to add to the previous issue)

The result of the global data check may be a failure. This is not taken into account for (2); I presume that, in the case of a failure, value[key] should be deleted in the first branch, and item[key] in the second.

index.html
mattgarrish: I did a couple of tweaks to this section in the last commit. In addition to adding the scheme step, I also put a first step in to check that both the incoming value and base are non-empty strings.

I don't think it's problematic to require base be a non-empty string, since manifests have to have a base. If they don't, the publication is invalid, although this step will have the unintended consequence of removing all URLs in such a scenario, even if they are absolute already.

mattgarrish:
> Are we sure that the current choices are the right ones?

But given that typing isn't required, how do you determine what kind of object you have when the type is not specified? Are you trying to guess based on properties?

mattgarrish:
We could inject types for every known value category first, but we'd need an earlier step for entities and linkedresources to do that.

iherman:
> But given that typing isn't required, how do you determine what kind of object you have when the type is not specified? Are you trying to guess based on properties?

By 'type' I mean Typescript type (more exactly, interface) not types per in the manifest. Sorry for the confusion. I realize this may be an implementation specificity, and maybe we should be bit more vague on what 'context' means in the document...

iherman:
Mini issue: in 4.3.2.2., (2) there is a missing ',' between 'context' and 'term'

mattgarrish:
I think it's safer to use the type of object to determine context than to pass around property names, the more I think about this, though.

For one, it means we don't have to descend into every map but can limit processing to linkedresource, person and organization. It's also simpler to process url with a context of 'linkedresource' rather than check if it's in one of readingOrder, resources, links, or alternate (long term maintenance gets reduced, too).

See if PR #129 makes sense to you.

iherman:
4.3.2.2/(2): Although the details seem to be fine but... when does this branch occur in the first place? By the time we get to this step, everything turns into arrays...

mattgarrish:
> By the time we get to this step, everything turns into arrays...

Not everything, no. abridged, datePublished, LinkedResource/url, etc. are not turned into arrays.

I don't think the map check is useful, to be honest, since we don't have anything that accepts just a single object, but seems harmless as far as future processing goes.

iherman:
> > By the time we get to this step, everything turns into arrays...
> Not everything, no. abridged, datePublished, LinkedResource/url, etc. are not turned into arrays.
>
> I don't think the map check is useful, to be honest, since we don't have anything that accepts just a single object, but seems harmless as far as future processing goes.

I am sorry, that is not what I meant. Of course you are right. But there are no 'maps' of accepted types as values at this point.

> I don't think the map check is useful, to be honest, since we don't have anything that accepts just a single object, but seems harmless as far as future processing goes.

It is a never-used-branch. But I can see your point as a future proof thing. Maybe it is worth putting this into the text as a note, because other implementers will ask themselves the same question.

iherman:
As for future proof: if we explicitly list the acceptable types, per your latest change, maybe that is a new extension point: extensions may introduce their own additional types!

mattgarrish:
>  But there are no 'maps' of accepted types as values at this point.

Yes, right, I think we're on the same page now. :) I thought you meant only the array step was relevant. All objects are arrays.

> Maybe it is worth putting this into the text as a note, because other implementers will ask themselves the same question.

Makes sense.

> As for future proof: if we explicitly list the acceptable types, per your latest change, maybe that is a new extension point: extensions may introduce their own additional types!

Yes, good point. Even more benefits to the change. I'll modify the definition to note that the list can be extended.

mattgarrish:
In the last commit I made a new section called "Processing Contexts" and added that the list is extensible, plus added a note to ensure that types get added if you do. (Like the error handling section, was clogging up the start of the algorithm.)

iherman:
In the latest incarnation of the algorithm: we say, for normalize data, that the context parameter is optional. I guess that the intention is that if it is optional, the context is the 'global' one, which is fine, but that is not said in the text...

mattgarrish:
>  I guess that the intention is that if it is optional, the context is the 'global' one, which is fine, but that is not said in the text...

Right, I note this in the new processing section but I added it in a secondary commit: https://github.com/w3c/pub-manifest/pull/128/commits/1f43190dff0f447285914a9e7c140e2025401e16

Or you can view it here: https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/128.html#processing-contexts

Not sure my "direct child of the manifest" is the clearest. Is "global context" the more accurate way of phrasing it?

iherman:
>  Is "global context" the more accurate way of phrasing it?

Yes.

mattgarrish:
> > Is "global context" the more accurate way of phrasing it?
> 
> Yes.

How about: "If a context is not provided to a function, the term being processed is considered part of the global context (i.e., it is a direct child of the manifest object)."

PR #130, opened 2019-10-21
Setup JSON Schema tests

BigBlueHat:
This is just the first pass at getting the foundational code in place. This isn't (yet) enough to handle proper per-feature test reporting.

To do per-feature test reporting we will indeed need to break up each schema into smaller ones similar to what @tcole3 did for the Web Annotation tests at https://github.com/w3c/web-annotation-tests/tree/master/annotations

The reason is because we need per-item *pass* statements--not just a single "valid" for the whole document.

If someone knows of more granular JSON Schema testing tools (or [ajv](https://github.com/epoberezkin/ajv#options) configuration/setups), I'd love to avoid the manual process of breaking out these tests...

However, right now that's the best way to get that level of granularity in the output reporting.


schema/url.schema.json
```diff
@@ -3,5 +3,5 @@
...
+    "format": "uri-reference"
```
mattgarrish: FYI, see https://github.com/w3c/pub-manifest/pull/85 for why we changed these to regex. In a nutshell, we support iri-reference) but only uri-reference works) with ajv, and neither limits the scheme to http(s). the regex solution isn't ideal, but with 2019-09 of json schema it looks like format is being discouraged altogether so this may have to migrate out to another validation layer anyway.

iherman:
I have no problem with the content of the PR, (although I admit I have never used mocha, so I do not know how exactly it works, but I trust @BigBlueHat on this), but I am a bit worried about the current form. If we merge it, we will have a repository as a mixture of a document-oriented repository where the specification is defined and developed, and a program development, node.js based repository, with package.json, node libraries, etc. My worry is that these may clash at some point (e.g., on future PR-s), and the whole repository becomes very complicated to manage. 

I would rather be in favor of setting up a _separate test repository_ for this and possibly other test suites and manage them separately. There is this test, there should be a reporting mechanism for the API test, for other tests, some sort of a reporting for all of them, etc. Github repositories are cheap, it is a matter of cca 5 minutes for me to set them up, and it may avoid headaches.

@BigBlueHat @mattgarrish @TzviyaSiegman @GarthConboy @wareid @llemeurfr ?

dauwhe:
> Github repositories are cheap, it is a matter of cca 5 minutes for me to set them up, and it may avoid headaches.

As the number of Github repositories approaches infinity, it becomes impossible to find the repository you want. Perhaps the publ-wg repo readme could serve as a guide to what content is where, and include links to all the different repos (there are 33 tagged "publishing"!)

iherman:
@dauwhe there is a link on the right-hand site of the the WG's home page, linking to:

https://github.com/search?q=topic%3Apubl-wg+org%3Aw3c&type=Repositories

mattgarrish:
> there is a link on the right-hand site of the the WG's home page

You can also click on the publ-wg topic at the top of any of the github repository home pages -- the publishingcg and sync-media-pub repositories are maybe mislabelled as wg repositories, though.

(What might be more useful for general readers would be a publishing tag for the [list of all W3C standards and drafts](https://www.w3.org/TR/), but let's not continue this in a PR.)

iherman:
Overtaken by events... and the creation of publ-tests.

PR #131, opened 2019-10-22
Add section on publication resources

mattgarrish:
As a possible (partial) solution to https://github.com/w3c/audiobooks/issues/26, this PR adds a section explaining publication resources and also adds an algorithm for getting the list from the internal representation.

The stuff about comparing URLs is dropped, as it doesn't seem particularly useful.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/131.html" title="Last updated on Nov 1, 2019, 1:03 PM UTC (01bc1f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/131/3f2e675...01bc1f1.html" title="Last updated on Nov 1, 2019, 1:03 PM UTC (01bc1f1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4089,15 +4089,37 @@
 					</li>
 					<li>
 						<p><a href="https://infra.spec.whatwg.org/#set-iterate">For each</a>
-							<var>resource</var> of <var>internal_rep["readingOrder"]</var>, <a
-								href="https://infra.spec.whatwg.org/#set-append">append</a>
-							<var>resource["url"]</var> to <var>resources</var>.</p>
+							<var>resource</var> of <var>internal_rep["readingOrder"]</var>:</p>
+						<ol>
+							<li>
+								<p><a href="https://infra.spec.whatwg.org/#set-append">append</a>
+									<var>resource["url"]</var> to <var>resources</var>.</p>
+							</li>
+							<li>
+								<p>if <var>resource["alternate"]</var> is set, <a
+										href="https://infra.spec.whatwg.org/#list-iterate">for each</a>
+									<var>alternate</var> of <var>resource["alternate"]</var>, <a
+										href="https://infra.spec.whatwg.org/#set-append">append</a>
+									<var>alternate["url"]</var> to <var>resources</var>.</p>
+							</li>
+						</ol>
 					</li>
 					<li>
 						<p><a href="https://infra.spec.whatwg.org/#set-iterate">For each</a>
-							<var>resource</var> of <var>internal_rep["resources"]</var>, <a
-								href="https://infra.spec.whatwg.org/#set-append">append</a>
-							<var>resource["url"]</var> to <var>resources</var>.</p>
+							<var>resource</var> of <var>internal_rep["resources"]</var>:</p>
+						<ol>
+							<li>
+								<p><a href="https://infra.spec.whatwg.org/#set-append">append</a>
+									<var>resource["url"]</var> to <var>resources</var>.</p>
+							</li>
+							<li>
+								<p>if <var>resource["alternate"]</var> is set, <a
+										href="https://infra.spec.whatwg.org/#list-iterate">for each</a>
+									<var>alternate</var> of <var>resource["alternate"]</var>, <a
+										href="https://infra.spec.whatwg.org/#set-append">append</a>
+									<var>alternate["url"]</var> to <var>resources</var>.</p>
+							</li>
+						</ol>
 					</li>
 					<li>
 						<p>Return <var>resources</var>.</p>
--- a/index.html
+++ b/index.html
@@ -3053,42 +3053,6 @@
 						</details>
 					</li>
 
-					<li id="processing-unique-resources">
-						<p>Set <var>processed["uniqueResources"]</var> to the result of running <a>obtain a list of
-								unique resources</a> given <var>processed["readingOrder"]</var> and
-								<var>processed["resources"]</var>.</p>
-						<details>
-							<summary>Explanation</summary>
-							<p>This step compiles a list of all the unique resources within the <a
-									href="#publication resources">bounds of the publication</a>.</p>
-						</details>
-					</li>
-
-					<li id="processing-links">
-						<p>If <var>processed["links"]</var> is set, <a
-								href="https://infra.spec.whatwg.org/#list-iterate">for each</a>
-							<var>link</var> in <var>processed["links"]</var>:</p>
-						<ol>
-							<li>
-								<p>let <var>url</var> be the result of running <a
-										href="https://url.spec.whatwg.org/#concept-url-serializer">URL
-									serializer</a>&#160;[[!url]] on <var>link["url"]</var> with the <var>exclude
-										fragment flag</var> set.</p>
-							</li>
-							<li>
-								<p>if <var>processed["uniqueLinks"]</var>
-									<a href="https://infra.spec.whatwg.org/#list-contain">contains</a>
-									<var>url</var>, <a>validation error</a>, <a
-										href="https://infra.spec.whatwg.org/#list-remove">remove</a>
-									<var>link</var> from <var>processed["links"]</var>.</p>
-							</li>
-						</ol>
-						<p>If <var>processed["links"]</var> is an empty <a href="https://infra.spec.whatwg.org/#list"
-								>list</a> after processing, <a href="https://infra.spec.whatwg.org/#map-remove"
-								>remove</a>
-							<var>processed["links"]</var>.</p>
-					</li>
-
 					<li id="processing-end">
 						<p>Return <var>processed</var>.</p>
 					</li>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -2738,6 +2737,28 @@ <h4>Extensions</h4>
...
+				obtaining this list is described in <a href="#processing-pub-resources"></a>.</p>
+
+			<p>All other resources are outside the bounds of the digital publication (e.g., resources listed in the
+						<a><code>links</code></a> section and hyperlinks in the content to external resources on the
+				Web).</p>
+
+			<p>The only restriction this specification places on publication resources is that they be obtainable by a
+				URL that uses either the <code>http</code> or <code>https</code> scheme (see <a href="#value-url"></a>).
+					<a>Profiles</a> of this specification, however, MAY restrict both the content type and location of
+				resources.</p>
```
BigBlueHat: What about relative URLs?

I'm also concerned about restricting this to `http` and `https`. Imagine if this spec were older and restricted it to just `http` URLs...and then `https` came along...i.e. not sure restricting scheme--when we don't also restrict a protocol or intended API usage--makes much sense.

index.html
```diff
@@ -4053,6 +4074,34 @@ <h3>Add HTML Defaults</h3>
 					</ol>
 				</section>
 			</section>
+
+			<section id="processing-pub-resources">
+				<h3>Obtain a List of Publication Resources</h3>
```
BigBlueHat: This section (and it's siblings) define API like activities...but no API names. Their "return ____" statements don't say what they're returning into. Perhaps some text could be added to the parent section describing how all of these are populating a single object/representation?

index.html
```diff
@@ -4053,6 +4074,34 @@ <h3>Add HTML Defaults</h3>
...
+							<var>resource</var> of <var>internal_rep["readingOrder"]</var>, <a>append</a>
+							<var>resource["url"]</var> to <var>resources</var>.</p>
+					</li>
+					<li>
+						<p><a href="https://infra.spec.whatwg.org/#set-iterate">For each</a>
+							<var>resource</var> of <var>internal_rep["resources"]</var>, <a>append</a>
+							<var>resource["url"]</var> to <var>resources</var>.</p>
+					</li>
+					<li>
+						<p>Return <var>resources</var>.</p>
```
BigBlueHat: This would make the original `resources` list unknowable (per this description) from the internal representation. The internal representation would only ever have the union.

index.html
mattgarrish: Can you open a new issue for this? It's a fair question, but it's not new to this PR, so not really the place to take it up.

index.html
mattgarrish: We're not defining APIs, though, only the common steps for a user agent to internally process the information. That's generally how infra is used to define these algorithms.

For some other random samples, see:
- https://dom.spec.whatwg.org/#ordered-sets
- https://encoding.spec.whatwg.org/#concept-encoding-run
- https://fetch.spec.whatwg.org/#collect-an-http-quoted-string
- https://url.spec.whatwg.org/#concept-url-parser

index.html
mattgarrish: Right now it's not modifying the internal representation; it's simply a means of extracting the list from the two properties (the variable `resources` is not the same as the resource list).

If we want to add the result into the internal representation, then, yes, we'd need a different term to store it under to avoid clobbering the existing property that uses that name. I'm not sure if it's necessary to store the list, though, or if it's only something the user agent might want to generate when needed. I don't have a strong opinion on this list, so would be interested to hear what others think.

mattgarrish: If it helps, I switched the variable name to `publication_resources` to avoid confusion with the existing manifest term.

index.html
iherman: @BigBlueHat the process in the section is supposed to run _after_ the processing algorithm has run (it relies on the 'internal representation'). By that point the relative url-s expressed in the manifest have been turned into absolute ones. Ie, that part of the comment should be o.k.

However, I fully agree with the second comment. It actually may also create issues with off-line UA implementations which may, in URI terms, rely on `file:` URIs, in fact, because everything is within a package. So yes, it is clearly an issue to discuss and I agree with @mattgarrish that this should be the subject of a separate issue, and is orthogonal to this PR.

mattgarrish: > off-line UA implementations which may, in URI terms, rely on `file:` URIs

Right, I think the restriction to http(s) was heavily dependent on the "web" publications underpinning.

index.html
iherman: I would actually be in favor of adding this entry to the processing part as an extra term in the internal format, and refer to this algorithm at the very end of processing.

mattgarrish:
The last commit makes the changes listed in https://github.com/w3c/pub-manifest/issues/138#issuecomment-547381986

Note that it allows fragments in the reading order, but this should still be discussed with the WG.

I also added a `publicationResources` term to the internal representation and included a step to compile it (using URL serializer to weed out fragment identifiers).

mattgarrish:
On second thought, I've switched `publicationResources` to `uniqueResources` just to (hopefully) reduce any confusion about having a third list.

index.html
```diff
@@ -2094,9 +2095,6 @@ <h4>Resource Categorization Properties</h4>
 							>privacy policy</a>, and <a href="#structural-rel">structural resources</a> like the <a
 							href="#table-of-contents">table of contents</a>.</p>
 
-					<p>Note that a particular resource's <a>URL</a> MUST NOT appear in more than one of these lists, and
-						a URL MUST NOT be repeated within a list.</p>
-
 					<p>The manifest MUST NOT include a reference to itself within any of these lists.</p>
```
iherman: We have to be precise here... If we have a publication that has only one html file in the RO, and the manifest is embedded in that file, then this statement (though precise!) may be misunderstood. The point is that, in such a case, the manifest URL includes a fragment, and it is that _complete_ URL is what MUST NOT be in the RO. The fragment-less URL may.

It may be worth adding a short note to clarify this...

index.html
```diff
@@ -2238,13 +2238,15 @@ <h5>Resource List</h5>
 
 						<p>The order of items is <em>not significant</em>.</p>
 
-						<p>The URLs MUST NOT include fragment identifiers. It is RECOMMENDED to use
-								<code>LinkedResource</code> objects with their <code>encodingFormat</code> values
-							set.</p>
+						<p>To avoid conflicting information about a resource, a particular resource's <a>URL</a> SHOULD
```
iherman: Actually, note to myself: at the moment, the context file does turn resources and lists into (RDF) lists, ie, make the order significant. This should be changed (and I will)

index.html
```diff
@@ -3597,6 +3629,33 @@ <h3>Data Validation</h3>
 									error</a>, set to "<code>ltr</code>".</p>
 						</li>
 
+						<li id="validate-resources">
+							<p>(<a href="#resource-list"></a>) If <var>data["resources"]</var> is set, let
+									<var>uniqueURLs</var> be an empty <a
+									href="https://infra.spec.whatwg.org/#ordered-set">set</a>. <a
+									href="https://infra.spec.whatwg.org/#list-iterate">For each</a>
```
iherman: Shouldn't this say 'ordered set'? This is the term used elsewhere. In my reading, the infra standard term 'set' and 'ordered set' seem to be synonymous (the section head says 'set' and the content refers to ordered sets; I would think this is an editorial bug in the infra standard) but we should be consistent...

index.html
iherman: My mistake. The context file was o.k., there is nothing to change. What was (is) indeed buggy is my memory, but there is no way to change that:-(

index.html
```diff
@@ -4172,6 +4288,7 @@ <h4>The <code>PublicationManifest</code> Dictionary</h4>
     required sequence&lt;LinkedResource>    readingOrder;
              sequence&lt;LinkedResource>    resources;
              sequence&lt;LinkedResource>    links;
+             sequence&lt;LinkedResource>    uniqueResources;
 };
```
iherman: I believe `uniqueResources` should be a `sequence<DOMString>`.

index.html
mattgarrish: Although I sort of wonder why we care if the manifest references itself? It's not like a user agent will get caught in an infinite loop or anything. Maybe a more practical statement might be:

> It is not necessary to include a reference to the manifest in any of these lists.

Just as a contrived example, what if you wrote a guide to publication manifests and wanted to put the manifest json for the publication into the reading order?

iherman: >  …you wrote a guide to publication manifests and wanted to put the manifest json for the publication into the reading order?

I love this example:-) 

I agree, there is no need to put a "MUST NOT" for something that, for most of the cases, does not make sense anyway (i.e., why would an author do something like that)...

Nuke it!

mattgarrish: It also means we don't have to have a validation check against every URL in all three lists just to pull out a rather harmless reference.

iherman:
Editorial mini issues:

- the " obtain a list of publication resources" link in 5.4.(9) seems to be buggy; ReSpec underlines it and it is not an active link.
- the tabulation of the 'profile' entry in the WebIDL seems to be off

iherman:
In 2.8.2.1, the text now says:

> although profiles of this specification restrict both their use as well as what schemes and features are supported

shouldn't it say 'MAY restrict'?

iherman:
It does say, for lists, that:

> Resources listed in the links list MUST NOT be listed in the default reading order or resource list.

which I believe is correct. However, this is not checked in the processing steps. This may have to be done in a separate step (11) in data validation, with duplicate entries removed from links.

iherman:
On more item: in data validation step 10 we should also check that the URL does not have a fragment. The text says that items in resources _SHOULD NOT_ have a fragment...

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-11-04-pwg#section1-1).

- No actions or resolutions
Garth Conboy: <a href="https://github.com/w3c/pub-manifest/issues/138">https://github.com/w3c/pub-manifest/issues/138</a><br/>
Garth Conboy: <a href="https://github.com/w3c/pub-manifest/pull/131">https://github.com/w3c/pub-manifest/pull/131</a> (corresponding PR)<br/>
Matt Garrish: a number of issues came up last week… dauwhe had raised a couple of questions… going through algo to implement the spec<br/>
… the major ones in 138 are that we’re not going to have a MUST NOT on fragment ids on manifest level, but could be restricted at the profile level<br/>
… the other one (PR 131) also deals with publication of resources in reading order and resource list<br/>
… we had “MUST NOT” but we will leave it at “SHOULD NOT”… if it’s there, we won’t mess around with your reading order<br/>
… strongly advised that you don’t declare resources multiple times in reading order… leads to ambiguity<br/>
… we want to generally avoid that issue<br/>
… other than that, the changes were mostly editorial cleanup<br/>
… to recap: fragment ids are allowed, and you can declare multiple resources, and it will be a warning (still)<br/>
… we can discuss the resource issues now… i don’t think we need to restrict fragment ids at the publication level<br/>
Garth Conboy: nobody on the queue… ivan made an off comment on alternate issues… what did you mean?<br/>
Ivan Herman: we also had an issue in the last few days on what happens if you provide an alternate [?]<br/>
Matt Garrish: choosing alternates is issue 133… separate from 138<br/>
Garth Conboy: let’s stick with 138 for the moment<br/>
… nobody on the queue… dauwhe, you had chimed in on this issue. are you happy with where we are?<br/>
Dave Cramer: yes, i think allowing fragment ids is important, but i’m fine with the decision on duplicate resources… when we make restrictions I want us to be very conscious about what we’re restricting and why<br/>
</details>

PR #140, opened 2019-10-31
Additional pub resources/algorithm cleanup

### LINKED ISSUE #136: Missing context reference in 4.4.2.2. of the processing steps
In 4.2.2./1/ii when running (recursively) the verify value category function, that function requires a context value, which should be `item`

The same is true for 4.2.2./2/b, with `value` being the context.

mattgarrish:
These changes are in addition to the ones in PR #131. Breaking them out to make them easier to read, but will merge them before Monday's call.

- restore the restriction on duplicate resources in reading order, but only as recommendation so the entries are not removed;
- remove the restriction against fragment identifiers in the structural relations (toc, pagelist, cover);
- simplify the steps that reference all the unsupported data types for a value (simplifies to "if not a map ..." as the final step always expects a map);
- handle lists of URLs in the URL normalize step;
- warn but don't remove duplicate resources from reading order and resource list;
- remove links that specify relations that have to be in publication bounds from the links section;
- add item["type"] as the context when recursively calling the verify value function (fixes #136);
- add IDs to various steps that didn't have them for easier referencing;
- change use of "expected type" to "expected value category" to avoid any confusion with object types;
- move note that explanations are informative to the conformance section as it applies across the various algorithms;
- clarify the process for finding the toc with a fragment identifier - element referenced by a fragment identifier if it has role=doc-toc, otherwise first element with role=doc-toc, otherwise no machine toc;
- remove expectation of a `nav` element for the toc from the processing algorithm, as it's not a requirement.

I believe this addresses all the ongoing issues, but let me know if I've missed anything. It's getting to be quite the list. :)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/140.html" title="Last updated on Nov 1, 2019, 12:13 PM UTC (110c3f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/140/b76ef13...110c3f0.html" title="Last updated on Nov 1, 2019, 12:13 PM UTC (110c3f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2246,8 +2246,7 @@
 						<p>The order of items is <em>not significant</em>.</p>
 
 						<p>To avoid conflicting information about a resource, a particular resource's <a>URL</a> SHOULD
-							NOT be repeated within the resource list. Only the first declaration of a resource is
-							retained during <a href="#manifest-processing">processing of the manifest</a>.</p>
+							NOT be repeated within the resource list.</p>
 
 						<p>The URLs expressed in the resource list SHOULD NOT include fragment identifiers.</p>
 
@@ -4046,6 +4045,31 @@
 										<p>otherwise, <a href="https://infra.spec.whatwg.org/#set-append">append</a>
 											<var>url</var> to <var>uniqueURLs</var></p>
 									</li>
+									<li id="unique-alternate-parse">
+										<p>if <var>resource["alternate"]</var> is set, for each <var>alternate</var> of
+												<var>resource["alternate"]</var>:</p>
+										<ol>
+											<li id="alternate-parse-url">
+												<p>let <var>alt_url</var> be the result of running <a
+														href="https://url.spec.whatwg.org/#concept-url-serializer">URL
+														serializer</a>&#160;[[!url]] on <var>alternate["url"]</var> with
+														<var>exclude fragment flag</var> set.</p>
+											</li>
+
+											<li id="alternate-parse-test">
+												<p>if <var>uniqueURLs</var>
+													<a href="https://infra.spec.whatwg.org/#list-contain">contains</a>
+													<var>alt_url</var>, <a>validation error</a>.</p>
+											</li>
+
+											<li id="alternate-parse-append">
+												<p>otherwise, <a href="https://infra.spec.whatwg.org/#set-append"
+														>append</a>
+													<var>alt_url</var> to <var>uniqueURLs</var></p>
+											</li>
+										</ol>
+									</li>
+
 								</ol>
 							</li>
 
--- a/index.html
+++ b/index.html
@@ -3001,14 +3001,13 @@
 									href="https://infra.spec.whatwg.org/#string">strings</a>, then <a
 									href="https://infra.spec.whatwg.org/#iteration-break">break</a>.</li>
 						</ol>
-						<p>If the value of <var>lang</var> is neither an empty <a
-								href="https://infra.spec.whatwg.org/#string">string</a> nor a <a
-								href="https://tools.ietf.org/html/bcp47#section-2.2.9">well-formed</a>&#160;[[!bcp47]]
-							language tag, <a>validation error</a>, set <var>lang</var> to an empty string.</p>
-						<p>If the value of <var>dir</var> is neither an empty <a
-								href="https://infra.spec.whatwg.org/#string">string</a> nor one of the values
-								"<code>ltr</code>" or "<code>rtl</code>", <a>validation error</a>, set <var>dir</var> to
-							an empty string.</p>
+						<p>If <var>lang</var> is not an empty <a href="https://infra.spec.whatwg.org/#string">string</a>
+							and is not a <a href="https://tools.ietf.org/html/bcp47#section-2.2.9"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4610,9 +4634,23 @@ <h3>User Agent Processing</h3>
...
-						<code>nav</code> element in document order with the <code>role</code> attribute value
-						<code>doc-toc</code>. All explanations are <em>informative</em>.</p>
+				<p>The following algorithm MUST be applied to a walk of a DOM subtree rooted at the table of contents
+					element determined as follows:</p>
+
+				<ol>
+					<li>
+						<p>if the <code>LinkedResource</code> that identifies the table of contents includes a fragment
+							identifier in its URL, the element referenced by the fragment identifier provided the
+							element has the <code>role</code> attribute value <code>doc-toc</code>.</p>
```
iherman: I believe it should say:

> element referenced by the fragment identifier, provided the element has the...

(comma was missing between 'identifier' and 'provided'

index.html
```diff
@@ -3675,6 +3677,14 @@ <h3>Data Validation</h3>
 											href="https://infra.spec.whatwg.org/#list-remove">remove</a>
```
iherman: In step (2), it should be <var>data["uniqueResources"]</var>, and not <var>["uniqueLinks"]</var>

index.html
```diff
@@ -4001,76 +4017,84 @@ <h4>Verify Value Category</h4>
...
+											href="https://infra.spec.whatwg.org/#set-append">append</a>
 										<var>url</var> to <var>uniqueResources</var>.</p>
 								</li>
-								<li>
+								<li id="append-alternates">
 									<p>if <var>resource["alternate"]</var> is set, set <var>uniqueResources</var> to the
 										result of <a>append URL</a> given <var>resource["alternate"]</var> and
-										<var>uniqueResources</var>.</p>
+											<var>uniqueResources</var>.</p>
 								</li>
```
iherman: Not exactly... <var>resource["alternate"]</var> is an array of linked resources, ie, it is not a valid argument type to invoke `append URL`. Instead, the function must be invoked, recursively, on each element of <var>resource["alternate"]</var>.

(I should have checked this yesterday :-(

iherman:
> I still have to update my implementation, maybe some more issues will come up but, at first glance, I do not think so...

I was proven wrong :-)

mattgarrish:
> I was proven wrong :-)

Ya, always one more tweak to make. I also noticed there was a useless variable still defined in the append url function when it looked like we'd need to remove items based on reading order/resources.

Now on to update my code and hope nothing else comes up... :)

mattgarrish:
Sure enough, this looks grammatically incorrect...

> If the value of lang is neither an empty string nor a well-formed [bcp47] language tag, validation error

This would trigger an error if lang is non-empty and if it is not well-formed. Should be:

> If lang is not an empty string and is not a well-formed [bcp47] language tag, validation error

mattgarrish:
Fixed the neither/nor problem and also added "depending on context" to the normalization steps without that text. I know it only matters right now for arrays, but it's probably better long term to always look at context, plus it looks strange to have a context variable and then not use it in most steps.

mattgarrish:
I've also removed a redundancy in the validation steps.

Step 10 was checking the resources list for duplicate URLs, but then step 11 ends up doing the same thing for both the reading order and resource list.

I've done away with step 10 and simplified what was step 11, as there wasn't a need for two function calls.

iherman:
@mattgarrish however, you've removed an additional step: to add the URL-s in a possible `alternates` for the final bounds.

I think this step must be added to 5.4.2.3 after step (2).

iherman:
Another problem: the new setup in (10) treat `resources` and `readingOrder` exactly the same way. However, for `resources` we have this extra remark:

>  To avoid conflicting information about a resource, a particular resource's URL SHOULD NOT be repeated within the resource list. _Only the first declaration of a resource is retained during processing of the manifest._

(Emphasis is mine, it is not present for `readingOrder`)

The algorithm may be o.k., actually, because the only thing it _may_ want to say is that only the first URL is used for the bounds, but that is also true for the `readingOrder`.

I actually do believe that it is way cleaner if the two lists _are_ treated in a similar manner, otherwise, this will be a source of confusion. Maybe the simplest is to remove this extra sentence from `resources`

mattgarrish:
> however, you've removed an additional step: to add the URL-s in a possible `alternates` for the final bounds.

D'oh, I forgot to merge that from the original list. Thanks.

> I actually do believe that it is way cleaner if the two lists _are_ treated in a similar manner, otherwise, this will be a source of confusion. Maybe the simplest is to remove this extra sentence from `resources`

Yes, I don't think we should be removing entries from the reading order or resource list, only warning about them. The unique list of resources already takes care of removing any redundancy and giving precedence. I'll strip that sentence.

iherman:
@mattgarrish I think that, at this point, we should merge this into the other PR. It becomes too confusing to have both:-)

As far as I can see, we could merge that one, too, i.e., we are essentially done. We only need the WG's approval on Monday...

mattgarrish:
Ya, I was just leaving these open while I updated my code, but at this point I think I've run through all the major changes.

Issue #143, opened 2019-11-01
repeated structural relations


mattgarrish:
We don't restrict the use of structural relations (cover, contents and pagelist) to a single resource, only that they not occur in both the reading order and resource list.

This seems like an oversight, but if multiple are listed we should still make explicit that first found wins (and validate this).

iherman:
Alas! +1 to 

> This seems like an oversight, but if multiple are listed we should still make explicit that first found wins (and validate this).

Alas!, because it is an extra complication on the algorithm:-(

mattgarrish:
Actually, validating the toc and pagelist is quite easy (extend reading order with resource list and walk over each entry), but the cover is proving a bit of a nightmare, at least thinking about it practically.

To be more specific, though, we say that multiple covers are allowed provided they differ on at least one property. Checking each one is unique is going to get interesting. The only property we define that really could be looked at is `encodingFormat`, but it's kind of useless for the suggested reasons for allowing multiple covers. The spec cites uses such as height, width and rel (?), but we don't have properties for height and width and I don't know what different rel value would help.

We need media queries to make multiple covers useful, and we don't have them by default.

It also feels like if we're going to allow mutliple covers, we should also restrict the rel to one resource and use `alternate` to list the other options. We defined covers long before alternate came around, unfortunately, but this is kind of a perfect case for it.

At any rate, I'm going to open a PR in a second that strictly interprets the spec as it is right now, but I think we should modify cover alternates and I'm not sure exactly what we do about uniqueness. We can get away without being specific in the algorithm, but it feels underspecified.

mattgarrish:
Is there a way to have LinkedResource extend MediaObject through the context so that its properties are easily available without having to declare a type?

That would at least get us some useful properties for differentiating covers.

iherman:
> Is there a way to have LinkedResource extend MediaObject through the context so that its properties are easily available without having to declare a type?

"subtyping", in terms of RDF, is doable, but this is not something schema.org does, afaik; I do not think they formally do anything of the sort.

What *is* possible is to say that a cover page has two types, ie, it mixes in a media object. But that latter is an object with a large number of properties (it is also a creative work), ie, that would not help us too much.

[...]

PR #144, opened 2019-11-02
enforce rel requirements

mattgarrish:
This PR makes the following changes:

- it limits the contents and pagelist relations to one resource
- it adds a step to validate the relation requirements - also includes that covers should have a name and description
- I also moved the structural relations ahead of the informative relations in the spec, as I find it annoying the less important relations are listed first

But I think this needs more work as described in #143.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 3, 2019, 10:44 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fpub-manifest%2F3270e0d2e7300fa25afd949c636123d17de9114a%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/w3c/pub-manifest/3270e0d2e7300fa25afd949c636123d17de9114a/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/pub-manifest%23144.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3694,17 +3694,19 @@
 										false.</p>
 								</li>
 								<li id="rel-create-list">
-									<p>Let <var>resources</var> be the result of <a
-											href="https://infra.spec.whatwg.org/#list-extend">extending</a>
-										<var>data["readingOrder"]</var> with <var>var["resources"]</var>.</p>
+									<p>Set <var>resources</var> to the value of <var>data["readingOrder"]</var>, when
+										defined, otherwise to an empty <a
+											href="https://infra.spec.whatwg.org/#ordered-set">ordered set</a>. <a
+											href="https://infra.spec.whatwg.org/#list-extend">Extend</a>
+										<var>resources</var> with <var>var["resources"]</var>, when defined.</p>
 								</li>
 								<li id="rel-iterate-list">
 									<p><a href="https://infra.spec.whatwg.org/#list-iterate">For each</a>
-										<var>resource</var> of <var>resources</var>, if <var>resource["type"]</var> is
+										<var>resource</var> of <var>resources</var>, if <var>resource["rel"]</var> is
 										set:</p>
 									<ol>
 										<li id="rel-check-contents">
-											<p>if <var>resource["type"]</var>
+											<p>if <var>resource["rel"]</var>
 												<a href="https://infra.spec.whatwg.org/#list-contain">contains</a> the
 												value "<code>contents</code>":</p>
 											<ol>
@@ -3718,7 +3720,7 @@
 											</ol>
 										</li>
 										<li id="rel-check-pagelist">
-											<p>if <var>resource["type"]</var>
+											<p>if <var>resource["rel"]</var>
 												<a href="https://infra.spec.whatwg.org/#list-contain">contains</a> the
 												value "<code>pagelist</code>":</p>
 											<ol>
@@ -3732,7 +3734,7 @@
 											</ol>
 										</li>
 										<li id="rel-compile-covers">
-											<p>if <var>resource["type"]</var>
+											<p>if <var>resource["rel"]</var>
 												<a href="https://infra.spec.whatwg.org/#list-contain">contains</a> the
 												value "<code>cover</code>":</p>
 											<ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3678,6 +3685,90 @@ <h3>Data Validation</h3>
...
+								<li id="rel-booleans">
+									<p>Let <var>contents</var>, <var>pagelist</var> and <var>cover</var> be <a
+											href="https://infra.spec.whatwg.org/#boolean">boolean</a> values set to
+										false.</p>
+								</li>
+								<li id="rel-create-list">
+									<p>Let <var>resources</var> be the result of <a
+											href="https://infra.spec.whatwg.org/#list-extend">extending</a>
+										<var>data["readingOrder"]</var> with <var>var["resources"]</var>.</p>
+								</li>
```
iherman: The [infra "extending"](https://infra.spec.whatwg.org/#list-extend) operation says:

> To extend a list A with a list B, for each item of B, append item to A.

in human readable:-) term, this means `A = A append B`. 

In our case that can be misrepresented, and a false reading would yield an extended `data["readingOrder"]`. Clearly, any mindful implementer will realize that we do a run through the list without that assignment, we may want to be a bit more careful in the formulation.

index.html
```diff
@@ -3678,6 +3685,90 @@ <h3>Data Validation</h3>
...
+								<li id="rel-iterate-list">
+									<p><a href="https://infra.spec.whatwg.org/#list-iterate">For each</a>
+										<var>resource</var> of <var>resources</var>, if <var>resource["type"]</var> is
+										set:</p>
```
iherman: Here and in the rest of this step it should say <var>resource["rel"]</var> and not <var>resource["type"]</var> :-)

iherman:
Another slightly edge case (it is a matter of taste, really). It says now:

```
if resource["type"] contains the value "cover":
if cover is false, set cover to true and check:
    if resource["name"] is not set or is an empty string, validation error.
   if resource["description"] is not set or is an empty string, validation error.
otherwise, validation error.
```

I wonder whether it is not more logical to say that the 'cover' is set to true if that stuff is there _and_ it is valid. The algorithm, as it stands now, will flag a cover page as existing even if it is not valid even if there is another one down the line that is perfectly valid.

(I do not feel strong about this...)

index.html
mattgarrish: Sure, we could assign readingorder to resources first and then extend.

mattgarrish: Actually, this makes sense as the reading order and resource list might not be set, so it allows some default checking on both:

> Set `resources` to the value of `data["readingOrder"]`, when defined, otherwise to an empty ordered set. Extend `resources` with `var["resources"]`, when defined.

index.html
mattgarrish: wow, can't even blame that on a typo. my mind was elsewhere...

mattgarrish:
> I wonder whether it is not more logical to say that the 'cover' is set to true if that stuff is there _and_ it is valid.

I'm hesitant about this, as they're only warnings. If they were stronger requirements than "should", maybe, but otherwise we're sort of ignoring that you're not supposed to list these twice and that we give precedence to the first one found.

And what if you don't find one better specified, do you have to go back to the first, or is there a formula for which is best of not perfect options?

End of the day it's an authoring mistake to have two covers, and incomplete information is one of the things that can go wrong. Better to warn the author and have them correct it than the user agent spend time trying to solve this, IMO.

index.html
iherman: Which is exactly how I implemented it:-) 👍

PR #146, opened 2019-11-05
possible formulation of alternate

mattgarrish:
Here's an attempt to thread the needle on alternates.

It seems like we have to define at least minimal expectations if we want to still talk about alternate covers, or have alternates for audiobooks. To that end, I explain that we only define the property for use with `encodingFormat`. I've also defined a minimal selection algorithm.

Does this look like it will work?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/146.html" title="Last updated on Nov 5, 2019, 12:36 PM UTC (312950a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/146/e4a68d7...312950a.html" title="Last updated on Nov 5, 2019, 12:36 PM UTC (312950a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -797,7 +797,8 @@
 						<p>This specification only defines the <code>alternate</code> property for selecting from
 							alternative formats (i.e., based on <code>encodingFormat</code> or by inspecting URLs).
 								<a>Profiles</a> MAY extend this behaviour to allow selection based on other criteria.
-							The process for selecting an alternate is described in <a href="#select-alternate"></a>.</p>
+							The process for selecting an alternate is described in <a href="#app-select-alternate"
+							></a>.</p>
 
 						<p class="note">When defining a <code>LinkedResource</code> object, it is advised to always
 							specify the media type of the resource using the <code>encodingFormat</code> property. Doing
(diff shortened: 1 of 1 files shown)


iherman:
There was a temporary breakdown of preview; for this PR, the intermediate solution is to use:

Preview: https://raw.githack.com/w3c/pub-manifest/editorial/clarify-alternate/index.html

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fpub-manifest%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fpub-manifest%2Feditorial%2Fclarify-alternate%2Findex.html

iherman:
I am not 100% sure that the new 5.5 should really be present in section 5. It is a bit unrelated to the rest of that section, and is more in line with the appendix on TOC. I think it is more appropriate to have it in its own Appendix.

iherman:
Should we also add that future revisions of this specification may provide a more formal description of the choice? I am not sure at all, but it was mentioned on the call.

mattgarrish:
> I am not 100% sure that the new 5.5 should really be present in section 5.

Ya, I flip-flopped a couple of times. I started it in 5.5 just because of the infra aspect, moved it to an appendix and then moved it back. I'm fine either way.

mattgarrish:
> Should we also add that future revisions of this specification may provide a more formal description of the choice?

I'm kind of wary of these sorts of statements now. We've tried this in EPUB and they sort of linger around forever while nothing happens. It's always true that we could better define any part of the specification in the future.

PR #148, opened 2019-11-05
Revise alt text/extended description requirement

### LINKED ISSUE #145: cover description always recommended?
I noticed that we recommend both `title` (this is a mistake I'll fix, as it should be `name`) and `description` for all cover images.

Recommending `name` for alt text seems fine, but not every cover needs a description, only ones with complex images. Can we reword this more along the lines of:

> If the cover is in an image format, a `name` SHOULD be provided. User agents can use this property to provide alternative text for the image. The `description` property can also be used to provide an extended description, when necessary.

/cc @GeorgeKerscher @avneeshsingh @clapierre

mattgarrish:
This PR fixes #145 as follows:

- it replaces the recommendation to always include both a name and description with a pointer to WCAG SC 1.1.1 for describing images
- it adds a paragraph on generating alt text when it isn't specified
- it moves the paragraphs in the cover section around a bit so they match up with their relevant examples

For generating alt text, I kept the wording very simple like we've done with other cases where explain how to generate missing values. A user agent isn't required to follow the method we suggest, but we offer up a reliable option. I also avoided suggesting specific wording to avoid i18n issues.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/148.html" title="Last updated on Nov 5, 2019, 4:26 PM UTC (aee8a07)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/148/e4a68d7...aee8a07.html" title="Last updated on Nov 5, 2019, 4:26 PM UTC (aee8a07)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2492,14 +2492,6 @@
 						<p class="ednote">The <code>cover</code> term is not currently registered in the IANA link
 							relations but the Working Group expects to add it.</p>
 
-						<p>If the cover is in an image format, a <code>title</code> and <code>description</code> SHOULD
-							be provided. User agents can use these properties to provide alternative text and
-							descriptions when necessary for accessibility.</p>
-
-						<p>Only one resource MAY be identified as the cover, but additional covers MAY specified using
-							the <a><code>alternate</code></a> property (e.g., to provide alternative dimensions or
-							resolution).</p>
-
 						<pre class="example" title="Identifying an HTML cover page.">{
     …
     "resources" : [
@@ -2513,6 +2505,16 @@
     ],
     …
 }</pre>
+
+						<p>If the cover is an image (whether embedded in an HTML resource or not), it is strongly
+							advised to follow <a href="https://www.w3.org/TR/WCAG/#non-text-content">Success Criterion
+								1.1.1</a>&#160;[[WCAG]] for the provision of alternative text and extended descriptions.
+							For image formats that do not provide the ability to embed this information, the <a
+								href="#linkedresource-name"><code>name</code></a> and <a
+								href="#linkedresource-description"><code>description</code></a> properties of
+									<a><code>LinkedResource</code></a> can be used to provide alternative text and
+							extended descriptions, respectively.</p>
+
 
 						<pre class="example" title="Identifying a cover image. Alternative text and a description are provided in the name and description properties, respectively.">{
     …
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2530,6 +2532,17 @@ <h5>Cover</h5>
     …
 }</pre>
 
+						<p>If a user agent requires alternative text for a cover image to make an interface accessible,
+							and the <code>name</code> property is not specified, it MAY attempt to construct the
+							alternative text from the publication metadata. This specification does not mandate how such
+							alternative text is created, but constructing a string that identifies that the image is the
```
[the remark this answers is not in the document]
mattgarrish: I've reformulated it slightly just to get rid of the reliability stuff. Not necessary.

index.html
```diff
@@ -2538,6 +2544,32 @@ <h5>Cover</h5>
...
+							and the <code>name</code> property is not specified, it MAY attempt to construct the
+							alternative text from the publication metadata. This specification does not mandate how such
+							alternative text is created. One method is to construct the alternative text as a string
+							that identifies that the image as the cover, followed by the <a href="#pub-title"
+								>publication title</a>.</p>
+
+						<p>Only one resource MAY be identified as the cover, but additional covers MAY specified using
+							the <a><code>alternate</code></a> property (e.g., to provide alternative dimensions or
+							resolution).</p>
+
```
iherman: This sentence repeats the one right before example 37. Only one of them should remain; my guess is that is a better place for it, and the one above should be removed.

index.html
```diff
@@ -3749,17 +3781,8 @@ <h3>Data Validation</h3>
...
-															<p>if <var>resource["name"]</var> is not set or is an empty
-															string, <a>validation error</a>.</p>
-														</li>
-														<li id="rel-cover-desc">
-															<p>if <var>resource["description"]</var> is not set or is an
-															empty string, <a>validation error</a>.</p>
-														</li>
-													</ol>
+														to <code>true</code>. If <var>resource["name"]</var> is not set,
+															<a>validation error</a>.</p>
```
iherman: The problem here is that the new text talks about the <var>name</var> requirement only for images:

> If the cover is an image (whether embedded in an HTML resource or not), it is strongly advised to follow Success Criterion 1.1.1   [  WCAG  ] for the provision of alternative text and extended descriptions. For image formats that do not provide the ability to embed this information, the  name  and  description  properties of  LinkedResource  can be used to provide alternative text and extended descriptions, respectively. In these cases, the  name  property  SHOULD  always be set — the property can be left empty for decorative images. 

while this check looks for <var>name</var> in all cases. Are we sure that any cover is an image? If so, the sentence above should be reformulated. If not, the check should be more complex, because it should then require a check on whether the URL is an image. And that is not obvious:

- either we require that the media type is set to an image (which may mean to list some media types that we consider as an image; can be done if we are careful to mention the fact that the media types can be extended in future) or
- the UA agent has to look into the file and find some ways of checking whether it is an image.

I am not sure which way we want to go. Checking the presence of <var>name</var> is a good thing, but making a complete check seems be fairly complex...

index.html
mattgarrish: Weird. I thought I corrected that in the merge conflict.

index.html
mattgarrish: Ya, my mind is toast right now. I meant to add an image requirement. Honest! :)

My thought is we limit ourselves to `encodingFormat`:

> If `encodingFormat` is set and specifies an image media type (`image/*`), and resource["name"] is not set, validation error.

There are various ways to work around this rule if you really want to, so I'm not sure it's necessary to dig in to file contents. The best enforcement is failing an accessibility standard, like WCAG, for those who have to take this requirement seriously, but that happens in a separate evaluation.

iherman: Yeah, that sounds like right. It is not bulletproof, but it catches most of the possible situations and it takes care of the honest mistakes...

PR #154, opened 2019-11-08
recommended practices review

mattgarrish:
I went back over the specification to double-check recommended practices (not currently caught in the reqlist generator) and found some problematic statements that I've removed/reworded to not be normative, and a few things not accounted for in the algorithm.

- Recommending to use linked records rather than additional properties is probably a bad idea. Technically, it means we should warn about every property not defined by the specification. It also picks a winner when we don't know all the real-world use cases for metadata. For those reasons, I'm suggesting we drop that recommendation with this PR.
- It's pointless to recommend using linked resources for linked records when linked resources are required by the structural properties.
- We recommend the `describedby` relation for descriptive metadata, but what constitutes "descriptive" metadata? I'm switching this to an example of how to add metadata.
- Having a must for metadata resources to set their media type conflicts with the general advisement to do this and necessitates figuring out which resources are metadata resources, which will be an inexact science, at best.
- It's not necessary to say linked records must be in the resource list when they are resources otherwise in the links. That's the meaning of those lists and there's no way to know if the record is in the right list or not.
- We recommend to to use public vocabularies for metadata properties, but there's no way to enforce this (plus what is "public"?).
- There's no need to say that linked records should indicate a rel value when that's already a general requirement for links.
- I've clarified why we recommend including the accessibility and privacy reports as publication resources.
- The recommendation to use a human-readable format for accessibility report is ambiguous and open-ended. I've changed this to an explanation of why human readability is important (users).
- Recommending embedded metadata in the accessibility report is going too far out into authoring restrictions, and could never be checked unless we restrict the reporting to html/schema.org. Removing.
- Adds rel checks (IANA keyword or URL) to the algorithm.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/154.html" title="Last updated on Nov 8, 2019, 5:37 PM UTC (78ff429)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/154/0d4a8f7...78ff429.html" title="Last updated on Nov 8, 2019, 5:37 PM UTC (78ff429)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -714,10 +714,7 @@
 									</td>
 									<td>The relation of the resource to the publication. OPTIONAL.</td>
 									<td>
-										<p>One or more <a href="#manifest-rel">relations</a>. The values are either the
-											relevant relation terms of the IANA link
-											registry&#160;[[!iana-link-relations]], or valid URL strings&#160;[[!url]]
-											if no suitable link registry item exists.</p>
+										<p>One or more <a href="#manifest-rel">relations</a>.</p>
 									</td>
 									<td>
 										<a href="#value-array">Array</a> of <a href="#value-literal">Literals</a>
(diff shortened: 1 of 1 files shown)


mattgarrish:
The rel restrictions are also problematic. In the attribute definition we say you should use a known vocabulary, but in the extensions we say IANA.

The problem with restricting ourselves to IANA is that there are a lot of relations in the microformats wiki that people might use, but would be invalid.

I've removed the duplicate definitions (only the extension section is needed) and removed the restriction to IANA. A consequence is that we probably can't validate rel, but I don't think this will prove a huge loss since we don't require any specific behaviours.

PR #162, opened 2019-11-11
Rearrange introduction sections

mattgarrish:
This PR moves the introduction that was in the pub manifest section up to the top-level introduction. We originally had an intro per section in wpub when we had three separate parts, but it seems redundant now.

Some of the intro sections have normative statements in them, so I'd still apply the non-normative label on a section-by-section basis.

The terminology and conformance sections are moved out of the intro, as this looks like a common practice in some other specs I looked at.

No prose has been changed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/162.html" title="Last updated on Nov 11, 2019, 5:00 PM UTC (6deb283)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/162/4f4a366...6deb283.html" title="Last updated on Nov 11, 2019, 5:00 PM UTC (6deb283)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -184,7 +184,7 @@
 					be able to read the manifest's specific shape and internalize the data.</p>
 			</section>
 
-			<section id="manifest-schemaorg">
+			<section id="manifest-schemaorg" class="informative">
 				<h4>Relationship to Schema.org</h4>
 
 				<p>Manifest properties, in particular those categorized as <a href="#descriptive-properties">descriptive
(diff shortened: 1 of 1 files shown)


iherman:
Shouldn't the whole of §1 be informative? Only §1.2 is marked as such; §1.1, §1.3, and §1.4 should also marked as such...

mattgarrish:
How can a scope be non-normative, though? It can't be just an informative description of a scope.

1.3 and 1.4 have normative statements in them, which rules out their being informative. The MAY in 1.4 is pretty redundant with the extension section, though, so that could be written out and made informative.

I don't see that introductions are required to be informative. The SVG intro is all normative and the HTML is mostly informative (but each informative subsection is marked):

- https://html.spec.whatwg.org/multipage/introduction.html#introduction
- https://www.w3.org/TR/SVG11/intro.html

respec also isn't capable of overriding an informative designation (there's no "normative" class).

PR #164, opened 2019-11-13
switch length to duration for LinkedResource

mattgarrish:
Harmonizes so we don't have two properties for the same purpose with different names and different valid values.

https://github.com/w3c/audiobooks/issues/42

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/164.html" title="Last updated on Nov 13, 2019, 12:51 AM UTC (35cd46f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/164/59baa51...35cd46f.html" title="Last updated on Nov 13, 2019, 12:51 AM UTC (35cd46f)">Diff</a>

### DIFF
--- a/experiments/audiobook/flatland-canonical.json
+++ b/experiments/audiobook/flatland-canonical.json
@@ -66,7 +66,7 @@
     {
       "url": "http://www.archive.org/download/flatland_rg_librivox/flatland_1_abbott.mp3", 
       "encodingFormat": "audio/mpeg", 
-      "length": 1371,
+      "duration": "PT1371S",
       "name": [
         {
           "value": "Part 1, Sections 1 - 3",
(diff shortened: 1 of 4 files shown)


mattgarrish:
Crazy thought for the day ... any reason why we need to allow _any_ duration value?

It would be so much simpler if we restricted durations to being expressed in seconds, with a decimal fraction, for example (e.g., when it comes time to add durations).

Would still conform to the schema.org definition without being overly complex.

iherman:
> It would be so much simpler if we restricted durations to being expressed in seconds, with a decimal fraction, for example (e.g., when it comes time to add durations).
> 
> Would still conform to the schema.org definition without being overly complex.

I am bit wary of restricting the range of possible values of an existing schema.org property. We use schema.org's duration (which is great), I believe we should just take over their value restriction (which is the ISO format).

mattgarrish:
Seems like an ugly proposition just to maintain a bit of technical purity.

The result is we'll probably have to convert all durations to seconds, anyway, in order to add and compare them as part of the audiobook requirements.

PR #165, opened 2019-11-13
Prioritize extension validation

mattgarrish:
Whether or not https://github.com/w3c/audiobooks/issues/48 results in a change, profile validation/defaults should be prioritized over the general manifest rules.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/165.html" title="Last updated on Nov 13, 2019, 4:16 PM UTC (aaa8873)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/165/f51698d...aaa8873.html" title="Last updated on Nov 13, 2019, 4:16 PM UTC (aaa8873)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3558,6 +3558,16 @@
 							</details>
 						</li>
 
+						<li id="validate-extension">
+							<p>If a <a>profile</a> specifies data validation checks, those steps are executed at this
+								point.</p>
+							<details>
+								<summary>Explanation</summary>
+								<p>Profile validation steps are prioritized over the default steps so that if profiles
+									have, for example, different default values to apply, those values get applied.</p>
+							</details>
+						</li>
+
 						<li id="validate-type">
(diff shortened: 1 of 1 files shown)


iherman:
What is the reason? For all other extension points the callout to the extension happens at the end of the process, either as the very last step or right before the empty array removal. I am wary of introducing an inconsistency...

mattgarrish:
Because validation adds warnings/values specific to profiles that can conflict with the base, while the others don't. It'll be unclear what was authored or inserted if profiles rules come later.

Take the recommended properties in audiobooks, for example. How can you validate whether readingProgression was authored if default values were inserted before you can check the profile rules? Same for type.

iherman:
@mattgarrish o.k., I see your point. +1

Issue #179, opened 2019-12-14
toc algorithm fixes and clarifications


mattgarrish:
Some more issues with the toc algorithm that need fixing based on discussions @iherman and I have been having:

- toc.entries is not set to null after processing the first list in the root element, which allows other lists to be joined. Step 3 needs to set the value to null if branches is empty (i.e., the root list is finished).
- the tests in 5.2 to exclude branches without a name or any entries also needs to be applied before adding the item to the root toc.entries in 5.3. As it is, the root can contain completely empty entries if a list item doesn't have a link or a sublist in it.
- step 6.2.2 refers to checking the reading order and resources, which predates our adding the compiled uniqueResources list to the internal representation. We should make reference to this processed list instead.
- the algorithm should return toc or null at the end. It sort of ends abruptly by only stating that a toc without entries is unusable
- some statements are type heavy (e.g., "the map in current_toc_branch") and could be made more readable by omitting, since there's no need to repeat the type of value in the variable
- step 6.2 has some incorrectly formatted variable names
- the algorithm isn't purely event driven, as it's possible to exit an element when walking the dom without first descending through it. We should consider adding a clarifying note
- example 71 says both entries are in current_toc_branch prior to merging, but the first is in the stack

iherman:
To make it clear: these issues came up while implementing the algorithm.

Issue #177, opened 2019-12-12
split toc a element relations into a list


mattgarrish:
We should probably state to split on whitespace and create a list from the attribute when processing since `rel` can contain 0..n relations (even if there are few useful rel values for a link in a table of contents).

Right now we just say to trim the value and copy it over, so any values are still concatenated together as a string.

iherman:
You mean in 6.2.4 ? Yep.

iherman:
(I may have to take a deep breath and implement the TOC extraction, too:-)

PR #180, opened 2019-12-14
Clarifications to the table of contents algorithm

mattgarrish:
This PR addresses the issues noted in #179 as well as the rel tokenizing issues in #177.

@iherman I'm not sure if the note about the event model captures what you wanted or not, so feel free to modify if not.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/180.html" title="Last updated on Jan 27, 2020, 2:16 PM UTC (910b47c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/180/a6b77a8...910b47c.html" title="Last updated on Jan 27, 2020, 2:16 PM UTC (910b47c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4960,7 +4960,7 @@ dictionary LocalizableString {
 											<a href="https://infra.spec.whatwg.org/#ordered-map">map</a>:</strong>
 								</p>
 								<pre>«[
-    "name" → "",
+    "name" → null,
     "url"  → null,
     "type" → null,
     "rel"  → null,
@@ -5078,8 +5078,8 @@ dictionary LocalizableString {
 								<p>Run these steps:</p>
 								<ol>
 									<li>
-										<p>If <var>current_toc_branch["name"]</var> is not an empty <a
-												href="https://infra.spec.whatwg.org/#string">string</a>, do nothing.</p>
+										<p>If <var>current_toc_branch["name"]</var> is not <a
+												href="https://infra.spec.whatwg.org/#nulls">null</a>, do nothing.</p>
 									</li>
 									<li>
 										<p>Otherwise:</p>
@@ -5096,11 +5096,6 @@ dictionary LocalizableString {
 															>accessible name</a>&#160;[[accname-1.1]] of the
 														element).</li>
 												</ul>
-												<p>If the resulting value of <var>current_toc_branch["name"]</var> is an
-													empty string (e.g., after removing any presentational elements and
-													trimming all leading and trailing whitespace), set
-														<var>current_toc_branch["name"]</var> property to <a
-														href="https://infra.spec.whatwg.org/#nulls">null</a>.</p>
 											</li>
 											<li>If the element has an <code>href</code> attribute and the URL in the
 												attribute resolves to a resource in <a href="#reslist-end"
--- a/index.html
+++ b/index.html
@@ -4741,9 +4741,13 @@ dictionary LocalizableString {
 					some steps, user agents are provided a choice in how to process the content to provide flexibility
 					for different presentation models.</p>
 
-				<p class="note">This algorithm is not defined in purely event driven terms, as inspecting all descendant nodes is not always necessary to obtain the needed information from the DOM. In some cases, an element, and all its descendants, is omitted immediately after it is processed on <em>enter</em>. An event approach could be applied, but would require modifying the algorithm to process/ignore the omitted nodes.</p>
-	
-	
+				<p class="note">This algorithm is not defined in purely event driven terms, as inspecting all descendant
+					nodes is not always necessary to obtain the needed information from the DOM. In some cases, an
+					element, and all its descendants, is omitted immediately after it is processed on <em>enter</em>. An
+					event approach could be applied, but would require modifying the algorithm to process/ignore the
+					omitted nodes.</p>
+
+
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5101,21 +5112,23 @@ <h3>User Agent Processing</h3>
 													href="https://infra.spec.whatwg.org/#nulls">null</a>.</li>
 											<li>If the element has a <code>rel</code> attribute, and the value of the
 												attribute is not an empty string after trimming leading and trailing
-												white space, set <var>current_toc_branch["rel"]</var> to the trimmed
-												value. Otherwise, set <var>current_toc_branch["rel"]</var> to <a
+												white space, split the trimmed value on whitespace and set
+													<var>current_toc_branch["rel"]</var> to the resulting <a
+													href="https://infra.spec.whatwg.org/#list">list</a> of tokens.
+												Otherwise, set <var>current_toc_branch["rel"]</var> to <a
```
iherman: Github does not let me comment on unchanged lines; but it is a consequence of this change: In step (4), where `current_toc_branch` is initialized, I would expect the `rel` value to be set to either an empty array or, most probably, to `null`.

index.html
```diff
@@ -4769,6 +4769,11 @@ <h3>User Agent Processing</h3>
 				<p>If a table of contents element is not found, the publication does not have a table of contents that
 					can be used for machine rendering purposes.</p>
 
+				<p class="note">This algorithm is not defined in purely event driven terms, as inspecting all descendant
+					nodes is not necessary to obtain the needed information from the DOM. In some cases, an element is
+					exited immediately after it is processed. An event approach could be applied, but would require
+					modifying the algorithm to process/ignore the omitted nodes.</p>
+
 				<ol>
```
iherman: (1) I think this note should come after the first paragraph. Otherwise, I believe the text is o.k.

(2) The usual terminology in the algorithm is 

> exit the element and continue processing with the next element.

Maybe a slightly better way is to emphasize the situation, something like

> exit the element without descending the sub-tree at this element and continue processing with the next element.

A bit more verbose but may help making things clearer...

iherman:
Forgot in the formal github review part: maybe adding a reference to this PR in the change log is also appropriate.

iherman:
A purely stylistic/editorial comment. I found the name `current_toc_branch` somehow misleading. The main function of `current_toc_branch` is to be the current toc entry (the url, the rel values, and all that jazz), and yes, it may have some deeper level toc attached to it, but that is not really a 'branch'. Maybe `current_toc_entry` or `current_toc_node` is better?

iherman:
Actually... I found a bug in my code, and I realized that it was due to my misunderstanding of the text: the use of the word 'exit'.

(1) On the one hand, we use the term 'exiting an XXX element' to denote the (pseudo) event of, well, exiting the element.
(2) On the other hand, we use the phrase 'Exit the element and continue processing with the next element.'

So... does (2) mean that:

(i) the element also goes through the 'exit event', ie, executes the relevant 'event handler'
(ii) the element is essentially forgotten and the next element is handled

If you think indeed in terms of events, the answer is (i). But I believe what we really mean in the algorithm (ii).

My interpretation of the spec code was (i). Specifically, in step 4.2.2. there is an 'exit' of the sort described in (2), which (erroneously) led to the execution of 4.3 and, in the test` c2.branches.08` that led to an exception in my code. Indeed, 4.3 turned `toc.entries` to `null`, which led to an exception down the line. By changing the interpretation to (ii), things are fine.

Maybe we can avoid this misunderstanding if we use the term 'leaving', instead of 'exiting' in section 4 situation like (1). For example, 4.3. would say "When leaving a list element" instead of "When exiting a list element". And/or we can say "forget the element and continue processing with the next element" for the situation in (2).

WDYT?

iherman:
@mattgarrish I have a separate PR (#182) against this PR to handle these issues.

index.html
mattgarrish: It looks like this has already been fixed, as I see rel -> null in step 4.

I'm fine with that, but then `type` should probably be set to null as well for consistency. And in both cases, the "otherwise" statement at the end becomes redundant so can be removed (both values will already be null if their respective conditions to add a value aren't met).

mattgarrish: I've actually switched all the properties to null by default.

The only one that affects processing is `name`, but it took me a while to figure out what it was doing. Instead of starting it off as empty string and then setting it to null if a name is an empty string, we should start it off as null and just leave an empty string if that's the result.

Setting it to null conflicts with 5.2, which is where a placeholder gets inserted when the name is an empty string.

index.html
mattgarrish: This looks like it's good now, too. The only thing I did was s/omitted/skipped/ in the note to match the terminology from the other PR.

mattgarrish:
>  Maybe `current_toc_entry` or `current_toc_node` is better?

Ya, I like `current_toc_node` as it doesn't say anything about whether it's a leaf or branch. (It also avoids a bit of ugliness of having `current_toc_entry['entries']`.)

mattgarrish:
I think between the merged PRs and the changes I've just made everything has been captured now, but can you take another look @iherman ?

iherman:
@mattgarrish yep, it is a go as far as I am concerned!

Issue #188, opened 2020-01-27
drop requirement for https for schema.org?


mattgarrish:
I noticed this statement in the manifest context section:

> Although Schema.org is often referenced using the http URI scheme, the vocabulary is being migrated to use the secure https scheme as its default. **The use of https when referencing Schema.org in the manifest is REQUIRED by this specification.**

Does this really matter, and how are we supposed to enforce it? https is required in the context URL, so that's already covered, otherwise we don't reference schema.org URLs (except in the context document, which we also control).

Any reason not to drop it?

iherman:
Just commented in the PR #189

mattgarrish:
Closing with changes in #189.

PR #189, opened 2020-01-27
add extension step for profiles to extend the context

### LINKED ISSUE #186: An extra extension point may be needed? (Additional context files)
It is a possible that a particular profile would need an additional JSON-LD context file (although we do not have an example right now). Wouldn't it be necessary to add such an extension point in the processing steps?

The proper place would be after step (3) in [§7.4](https://www.w3.org/TR/pub-manifest/#processing-algorithm).

mattgarrish:
This PR fixes #186 and #188 as follows:

- notes that profiles may add additional context URLs
- removes the paragraph about requiring https for schema.org
- adds a general extension point for processing the context by a profile's rules

(I can restore the https bit if you think it's a blocker, but I just don't see the value in it.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/189.html" title="Last updated on Jan 28, 2020, 11:48 AM UTC (6f24fc8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/189/24ac983...6f24fc8.html" title="Last updated on Jan 28, 2020, 11:48 AM UTC (6f24fc8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -930,6 +930,11 @@
 					<li>the <dfn>publication context</dfn>: <code>https://www.w3.org/ns/pub-context</code></li>
 				</ol>
 
+				<p class="note">Although Schema.org is often referenced using the <code>http</code> URI scheme, <a
+						href="https://schema.org/docs/faq.html#19">the vocabulary is being migrated</a> to use the
+					secure <code>https</code> scheme as its default. As a result, only the <code>https</code> scheme is
+					recognized in the publication manifest context.</p>
+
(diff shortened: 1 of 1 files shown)


mattgarrish:
> The old note is probably too much, but to have a green 'note' warning people about this may be useful nevertheless.

Sure, that works for me. I've moved the paragraph up immediately after the list where we define the schema.org context and rewrote the last sentence to: "As a result, only the `https` scheme is recognized in the publication manifest context."

Issue #176, opened 2019-12-11
table of contents in a sectioning element


mattgarrish:
While we are clear that the first doc-toc is the toc when a fragment identifier isn't supplied, is this potentially problematic given that a sectioning element (article, aside, etc.) could have its own toc and occur first?

This probably falls into the highly unlikely scenarios and may not be worth worrying about, but just has me wondering. I'm not even sure there's a good solution without always requiring a fragment identifier, as the nav could be inside a section (i.e., we'd start to get into picking and choosing sectioning elements).

iherman:
Clearly, the 'first doc-toc' is brittle, and the better way of doing this is to use the fragment ID. We could add an editorial note accordingly. But, otherwise, I believe we should keep it at that...

mattgarrish:
Identifying the toc with 'doc-toc' seems a bit unnecessary when allowing fragment identifiers. I wonder if we should say that only matters when the toc isn't specifically identified by a fragment, with the note that it is always best to be specific?

Does the user agent reject an element with a matching ID if the element doesn't have the role on it, even if there is a processable list of links inside?

iherman:
I think at least a warning would be appropriate. As far as I understand, the presence of doc-toc has accessibility implications, and we should push for that...

mattgarrish:
Thinking about this again, I find it all a bit muddled.

While I don't like relying on finding an element with a role, are we causing other problems by allowing fragments?

Namely, how do you avoid resource duplication when a fragment might mean different things in different contexts. For example, say I want to link to my toc that's embedded in a bunch of front matter. Will the fragment be interpreted as where to start reading by the user agent unless I duplicate an entry so that the fragment-less one is in the reading order and the fragment-id'd one is in the resources? But if I do that, it causes duplication of resources so the entry that's referencing the toc gets tossed out during processing and is no longer available.

We might want to consider going back to only finding the toc by role and just live with the one in a million publications that can find a legitimate reason for having a toc for something else first.

iherman:
> We might want to consider going back to only finding the toc by role and just live with the one in a million publications that can find a legitimate reason for having a toc for something else first.

This may be the wisest approach after all. 

Also, there are three places where this whole issue appears: §4.8.1.3 and §C.1 for the manifest spec and §4.2 in the audiobook spec. None of these mention fragments, they only refer to the resource as a whole. I would think we can take out the fragment issue from our discussions...

mattgarrish:
Closing this with the changes in #192

PR #192, opened 2020-01-28
disallow fragments for identifying the toc

mattgarrish:
Due to a variety of problems that fragments present (conflicting purposes, inability to duplicate references, no fragments in resource list, etc., see #176), this PR reverts to our previous restriction against using fragments for identifying the table of contents.

Other than preserving some recent clarifications, the prose is set back to what we had in the 2019-09-11 draft.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/192.html" title="Last updated on Jan 28, 2020, 12:57 PM UTC (e946c7b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/192/04f4371...e946c7b.html" title="Last updated on Jan 28, 2020, 12:57 PM UTC (e946c7b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2597,7 +2597,9 @@
 							sections of a <a>digital publication</a>.</p>
 
 						<p>The <dfn data-lt="toc">table of contents</dfn> is identified by the <code>contents</code>
-							link relation&#160;[[!iana-link-relations]].</p>
+							link relation&#160;[[!iana-link-relations]]. The <a>URL</a> expressed in the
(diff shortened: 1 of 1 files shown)


mattgarrish:
Looking over the wording, though, I don't like the explicit restriction on fragment identifiers given that the fragment could be legitimate for other purposes. I've taken that out and instead noted that the toc is the first element with the role inside the resource, and given a pointer to the structuring section where that is made normative.

PR #197, opened 2020-02-05
clarify that rel values are case insensitive

### LINKED ISSUE #196: are rel values case sensitive?
I noticed that in the definition and algorithm we don't mention anything about case-insensitively matching rel values, but case insensitivity is the default for html (https://html.spec.whatwg.org/multipage/links.html#linkTypes).

Whether we want case sensitivity or not, we should make it clear which we expect in both places.

mattgarrish:
Fixes #196

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/197.html" title="Last updated on Feb 5, 2020, 11:48 PM UTC (2be8825)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/197/4d26f8b...2be8825.html" title="Last updated on Feb 5, 2020, 11:48 PM UTC (2be8825)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -706,6 +706,8 @@
 									<td>The relation of the resource to the publication. OPTIONAL.</td>
 									<td>
 										<p>One or more <a href="#manifest-rel">relations</a>.</p>
+										<p>Keywords are <a href="https://infra.spec.whatwg.org/#ascii-case-insensitive"
+												>ASCII case-insensitive</a> [[!infra]] and MUST be compared as such.</p>
 									</td>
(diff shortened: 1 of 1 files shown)


mattgarrish:
>  should be converted into lower case values

Ya, I started out that way but then got worried that it kind of breaks the describe-not-prescribe nature of the algorithm. You don't have to lowercase to get the right value, just as there's no one way to compare durations, for example.

PR #199, opened 2020-02-10
add back reading order requirements

### LINKED ISSUE #198: Lost requirement for reading order
In the steps for processing default values, the document that links to the manifest is [added to the reading order](https://w3c.github.io/pub-manifest/#processing-defaults-reading-order
) if there isn't a reading order.

It appears we lost the justification for this when we took the web publications section out, however, as these two statements used to be part of the specification:

> [The primary entry page] is an [HTML] resource that is returned when accessing the Web Publication's address, and MUST be included in either the default reading order or the resource list.

> The default reading order MAY be omitted when it only consists of the primary entry page. When the default reading order is absent, user agents MUST include an entry for the primary entry page when compiling the canonical manifest.

The current specification says to look at the discovery section to find the justification for adding the linking page to the reading order, but without those statements we have a logic hole.

To make the specification consistent again, we should incorporate these back into the current text.

(It would also make it unnecessary for audiobooks to have to check if document.url is in the unique resources, as this should be checked in the main algorithm.)

mattgarrish:
This PR fixes #198 by adding back the primary entry page requirements to be a resource and be added to the reading order when one isn't present so that the processing algorithm matches the prose again.

I made some minor changes to the original language, since the concept of the page that links to the manifest being called the "primary entry page" wasn't retained.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/199.html" title="Last updated on Feb 12, 2020, 2:22 PM UTC (65d9188)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/199/8d159b4...65d9188.html" title="Last updated on Feb 12, 2020, 2:22 PM UTC (65d9188)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4301,8 +4301,9 @@
 						</li>
 
 						<li id="processing-defaults-linking-resource">
-							<p>(<a href="#manifest-link"></a>) If <var>data["uniqueResource"]</var> does not <a
-									href="https://infra.spec.whatwg.org/#list-contain">contain</a>
+							<p>(<a href="#manifest-link"></a>) If <a data-cite="!dom#concept-document-url"
+										><var>document.URL</var></a> is set and <var>data["uniqueResource"]</var> does
+								not <a href="https://infra.spec.whatwg.org/#list-contain">contain</a>
 								<var>document.URL</var>, <a>validation error</a>.</p>
 							<details>
 								<summary>Explanation</summary>
--- a/index.html
+++ b/index.html
@@ -3081,7 +3081,12 @@
 								any invalid data is removed from the final representation.</p>
 						</details>
 					</li>
-
+					
+					<li id="processing-extension">
+						<p>If a <a>profile</a> specifies additional processing functions that need to be run, those
+							steps are executed at this point.</p>
+					</li>
+					
 					<li id="processing-defaults">
 						<p>Set <var>processed</var> to the result of running <a>add default values</a>, when successful,
 							given <var>processed</var> and <var>document</var>, when specified. Otherwise, terminate
@@ -3091,11 +3096,6 @@
 							<p>This step checks if any information missing from the manifest can be obtained from the
 								HTML document that links to the document, or from other sources.</p>
 						</details>
-					</li>
-
-					<li id="processing-extension">
-						<p>If a <a>profile</a> specifies additional processing functions that need to be run, those
-							steps are executed at this point.</p>
 					</li>
 
 					<li id="processing-end">
--- a/index.html
+++ b/index.html
@@ -3081,12 +3081,12 @@
 								any invalid data is removed from the final representation.</p>
 						</details>
 					</li>
-					
+
 					<li id="processing-extension">
 						<p>If a <a>profile</a> specifies additional processing functions that need to be run, those
 							steps are executed at this point.</p>
 					</li>
-					
+
 					<li id="processing-defaults">
 						<p>Set <var>processed</var> to the result of running <a>add default values</a>, when successful,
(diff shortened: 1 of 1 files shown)


mattgarrish:
FYI, I'll remove the duplicate check for the pep in the reading order from audiobooks when I reintroduce https://github.com/w3c/audiobooks/pull/73

mattgarrish:
> This all seems correct, but there is a somewhat awkward consequence.

Right, I don't see how we get around this. We had resolved the primary entry page had to be a resource and then lost the prose. I don't recall during audiobooks development that there was a discussion/decision to reverse this (but I could always be wrong). I expect we just based the tests on what we found in the document without realizing the prose was missing.

Does it make sense that a primary entry page doesn't have to be listed as a resource? The alternative is that instead of warning we add a similar rule to the reading order that if it is not listed in the unique resources we add it automatically to the resource list. I'm fine with that approach, but it adds a little bit extra to the specification that we hadn't defined before.

> However, if the author forgets to add a reading order, the algorithm will add the PEP URL to the reading order

Adding default values really should be the last thing that we do, as they should only go in after all other processing has been done and we're sure the values aren't present. If we simply run the extension steps before, then a fatal error occurs on the empty audio reading order. I've pushed that change.

iherman:
>  I'm fine with that approach, but it adds a little bit extra to the specification that we hadn't defined before.

which is, strictly speaking, something we should not do in CR phase. Maybe we should leave it at that, and the tests all have to be updated...

iherman:
Are you sure you made the right change in your latest commit? What I thought of doing that, in §7.4.3. the check on whether the document.URL is in unique resources must go _after_ step 3, ie, where the profile, possibly, add some defaults (and possibly would mistakenly add some resources to the wrong place.

That being said, the switch of the order in §7.4 makes sense, too.

mattgarrish:
> which is, strictly speaking, something we should not do in CR phase.

Right, but it is odd that if there is no reading order we automatically insert the document into that list, but if the document isn't in the resource list we issue a warning and don't insert it. Logically it seems like if we're doing one automatically then we should do the other, as inserting in the resources seems less problematic than inserting into the reading order.

I don't feel that strongly about it that if it is going to cause issue moving forward we need to resolve it now, though.

mattgarrish:
> must go _after_ step 3, ie, where the profile, possibly, add some defaults

To make sure I'm following, so a profile could insert the entry page into the resource list instead of having it done at this level?

wareid:
> > must go _after_ step 3, ie, where the profile, possibly, add some defaults
> 
> To make sure I'm following, so a profile could insert the entry page into the resource list instead of having it done at this level?

It seems to me for audiobooks this is the best method, to avoid the conflict we created in the readingOrder, as it can only contain audio resources. Keep the PEP in the resources instead.

iherman:
>>must go after step 3, ie, where the profile, possibly, add some defaults
>
> To make sure I'm following, so a profile could insert the entry page into the resource list instead of having it done at this level?

Yes, it is what I thought. Whether we want to add this to the audiobook profile: in theory, yes. But we may have to re-issue a CR, because this is a substantial change...

(which is o.k., we can do that, just knowing the consequences.)

mattgarrish:
Okay, I've moved it to the end of the section in the latest commit.

iherman:
Minor knit:

> (§ 6.1 Linking) If data["uniqueResource"] does not contain document.URL, validation error.

should be surrounded with the test whether document is defined (ie, there is a PEP in the first place).

mattgarrish:
Ah, right. I was thinking we didn't go into the defaults unless it was set, but the default might still come from the user agent as a last last case.

PR #214, opened 2020-04-20
explain author/creator equivalence problem

mattgarrish:
Here's a weak-kneed attempt to dodge the problem of author-creator equivalency in schema.org. It blames schema.org for the problem, prefers author and generally avoids saying anything normative.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/214.html" title="Last updated on Apr 30, 2020, 9:46 AM UTC (ea0a248)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/214/34cc0a4...ea0a248.html" title="Last updated on Apr 30, 2020, 9:46 AM UTC (ea0a248)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,55 +10,49 @@
 		<link href="https://unpkg.com/reqlist/lib/reqlist.css" class="removeOnSave" rel="stylesheet" type="text/css" />
 		<link href="common/css/common.css" rel="stylesheet" type="text/css" />
 		<script class="remove">
-			// <![CDATA[
+			//<![CDATA[
 			var respecConfig = {
-				preProcess: [
-					prepare_reqlist
-				],
-				postProcess: [
-					convert_dfn_to_link,
-					add_reqlist_button
-				],
+				preProcess:[
+				prepare_reqlist],
+				postProcess:[
+				convert_dfn_to_link,
+				add_reqlist_button],
 				wg: "Publishing Working Group",
 				specStatus: "ED",
 				shortName: "pub-manifest",
 				previousPublishDate: "2020-03-17",
 				previousMaturity: "CR",
 				implementationReportURI: "https://www.w3.org/publishing/groups/publ-wg/implementation/results.html",
-				crEnd : "2020-03-31",
+				crEnd: "2020-03-31",
 				edDraftURI: "https://w3c.github.io/pub-manifest/",
-				editors: [
-					{
-						"name": "Matt Garrish",
-						"company": "DAISY Consortium",
-						"companyURL": "https://daisy.org",
-						"w3cid": 51655
-					},
-					{
-						"name": "Ivan Herman",
-						"url": "https://www.w3.org/People/Ivan/",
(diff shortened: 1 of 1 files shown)


iherman:
@llemeurfr, to be very practical and procedural (sorry about that, but sometimes it is necessary...).

If we _remove_ a term from the list, this means that the change is _substantial_, which means that we have to go through a somewhat heavier process to republish the CR (transition request to the Director, etc., modification of the test suite, etc.). Obviously, if there is a technical _bug_ in the specification then we can and indeed should do this. 

Editorial changes, which is the case for this PR, can be done and re-published without further ado.

In any case, such a change should require a WG approval at this point.

(Personally, I am hesitant to do that for something which is not technically _wrong_, just ambiguous.)

Cc: @TzviyaSiegman @wareid @GarthConboy

llemeurfr:
I understand your concern. This is a CR, and will become a Rec after all feedback from integrators is processed. Some feedback is editorial (the easy one), some (like this one) is about the content of the spec itself; a bit more touchy. But if only lightweight feedback is treated, the spec will not reach its goal of interoperability. 
Could we put in a pile the non-editorial feedback and treat it in a batch before the Rec is released?

mattgarrish:
> i think that keeping `creator` in the list of recommended properties is adding fuzziness to the spec

It's not actually recommended, though. The list of recommended properties is here: https://www.w3.org/TR/pub-manifest/#manifest-requirements

The rest of the properties are the ones that we expect will be commonly used for publishing or by user agents. There's no restriction on the use of any other properties in schema.org, whether we list them or not. So dropping creator doesn't do anything in the end but add confusion about why it's not listed and what's to be done with it. I'd rather we said something than nothing.

The one bright side is that audiobooks has gone further and [recommended the author property](https://w3c.github.io/audiobooks/#audio-creators) so that gives some additional practical direction on which to use.

But @iherman is right that this is a notable flaw in the scheme to have two properties being synonymous. I'd rather go further and define how to combine them, but that also leads to some considerable normative changes both in the spec and in the algorithm. That's why I don't feel great about this, but can put it in the "can live with" bucket for this revision.

iherman:
> I understand your concern. This is a CR, and will become a Rec after all feedback from integrators is processed. Some feedback is editorial (the easy one), some (like this one) is about the content of the spec itself; a bit more touchy. But if only lightweight feedback is treated, the spec will not reach its goal of interoperability.
> Could we put in a pile the non-editorial feedback and treat it in a batch before the Rec is released?

That will put us in the obligation of re-issuing a CR minimally one month before we would go to PR. We have to discuss our timing to Rec anyway, and I would fine adding this to a batch of such changes _if we have one_. But I am a bit uneasy to complicate things by re-issuing a CR only for this reason, to be honest.

mattgarrish:
I'm going to merge this PR now. If there's anything we need to look at in the future, please open a new issue.

PR #220, opened 2020-06-30
fix schema expected data types

### LINKED ISSUE #218: Schema for accessibilityHazard should allow string
It only allows array: 
https://github.com/w3c/pub-manifest/blob/master/schema/publication.schema.json#L79

### LINKED ISSUE #219: Schema for accessModeSufficient should require an array of objects
Right now, it requires a string or array of strings:
https://github.com/w3c/pub-manifest/blob/master/schema/publication.schema.json#L59

But it looks like it should be an object or array of objects:
https://www.w3.org/TR/pub-manifest/#accessModeSufficient

mattgarrish:
Fixes #218 and fixes #219.

- Allows a single string value for `accessibilityHazard`
- Requires one or more ItemList objects for `accessModeSufficient`

### DIFF
--- a/schema/ItemList.schema.json
+++ b/schema/ItemList.schema.json
@@ -1,0 +1,38 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "ItemList.schema.json",
+    "title": "schema.org ItemList object",
+    "type": "object",
+    "properties": {
+        "type": {
+            "anyOf": [
+                {
+                    "type": "string",
+                    "const": "ItemList"
+                },
+                {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "contains": {
+                        "const": "ItemList"
+                    }
+                }
+            ]
+        },
+        "itemListElement": {
+            "type": [
+            	"array"
+            ],
+            "items": {
+                "type": "string"
+            },
+            "uniqueItems": true
+        }
+    },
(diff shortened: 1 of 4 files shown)


mattgarrish:
Hm, I standardized the indenting on tabs so the diff is showing all of publication.schema.json as changed, but only the two properties at lines 59 and 83 were actually modified.

The actual relevant mods are:

```
"accessModeSufficient": {
     "type": [
         "string",
         "array"
     ],
     "items": {
        "type": "string"
      },
     "uniqueItems": true
},
```

 to:

```
"accessModeSufficient": {
	"oneOf": [
		{
			"$ref": "ItemList.schema.json"
		},
		{
			"type": "array",
			"items": {
				"$ref": "ItemList.schema.json"
			},
			"uniqueItems": true
		}
	]
},
```

and:

```
 "accessibilityHazard": {
    "type": "array",
    "items": {
        "type": "string"
    },
    "uniqueItems": true
},
```

to:

```

"accessibilityHazard": {
	"type": [
		"string",
		"array"
	],
	"items": {
		"type": "string"
	},
	"uniqueItems": true
},
```

mattgarrish:
I also noticed that we've been using "anyOf" for the object type, but this should really be "oneof" as you either have a single string value or an array, not one or more of these. I don't think it was possible to make a value that could take advantage of the open choice, but its clearer to use the pattern we intend.

PR #224, opened 2020-07-02
Schema data type fixes

### LINKED ISSUE #221: schema: alternate should not be required to be an array
alternate should allow string, object, or array of strings or array of objects

### LINKED ISSUE #222: schema: validate @context for language and dir
Right now, no error is given for an invalid language or direction in the context property. E.g. this does not raise any errors:

```
"@context": [
        "https://schema.org",
        "https://www.w3.org/ns/pub-context",
        {
         "language": "mmm@@mmm",
         "direction": "xyz"
        }
    ],
```

### LINKED ISSUE #223: schema: url should be allowed to be an array
Right now, the top-level "url" property can only be a single string.

mattgarrish:
Fixes the following:

- adds validation for language and direction in the manifest @context
- allows more than one url for address
- fixes alternate so that mixes of strings and linkedresource objects are allowed in the array, as well as allow a single string or linkedresource object as the value;
- also atomizes some tests to reduce future maintenance in other schemas

Fixes #221, fixes #222, and fixes #223

### DIFF
--- a/schema/context.schema.json
+++ b/schema/context.schema.json
@@ -21,7 +21,8 @@
 				"properties": {
 					"language": {
 						"$ref": "bcp.schema.json"
-					}
+					},
+					"direction": false
 				},
 				"required": ["language"]
 			},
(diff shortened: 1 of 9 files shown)


marisademeglio:
Here FYI are the sample snippets I tried with the new `@context` schema:

[Accepted](https://gist.github.com/marisademeglio/f7f1d54bcb284f48f2473a84735ff645)
[Errors](https://gist.github.com/marisademeglio/1b9ba97709960222a4891824b58c262e)

I believe it's working as expected. The language value can be just about any single word but that seems to be a limitation (or feature) of the BCP regex. I tried another flavor of BCP regex (that I happened to be using in another project), and the results were the same.

mattgarrish:
> The language value can be just about any single word but that seems to be a limitation (or feature) of the BCP regex.

Right, I made a stink about the language property at one point -- whether it had to have valid language codes or only conform to the pattern. The prevailing opinion is that conforming to the pattern is sufficient, and that's all that the specification requires.

PR #225, opened 2020-07-02
Move schema components to a subdirectory

mattgarrish:
Since we're going to reuse the schema components for audiobooks, I was thinking it might make sense to separate them into their own subdirectory. That way there's no confusion about what needs to be copied over to make the audiobook schema work.

It probably also makes it easier to understand the role of the primary schema for anyone new to these.

WDYT?


marisademeglio:
Sorry I missed this - my test site was caching some stuff so when I reviewed these changes, it was using old files :/ 

It looks like `ref` values should be IDs, not filepaths. I couldn't get it to work with `ajv` with `./module` in the `ref` value, but as soon as I took it out, it was ok, even with the module files in their subdirectory.

mattgarrish:
They're supposed to be URIs. From the specification:

> The value of the "$ref" property MUST be a string which is a URI-Reference. Resolved against the current URI base, it produces the URI of the schema to apply.

I don't have any issues in oxygen with them, but haven't tried ajv.

marisademeglio:
Oh ok, maybe it's a problem with ajv. I was looking at [this](https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref), which made me think it was an ID value. But I see it takes a URI value.

mattgarrish:
Does it make a difference if you change $id to:

>	"$id": "https://raw.githubusercontent.com/w3c/pub-manifest/master/schema/publication.schema.json",

I remember we had problems with using an absolute URL early on because it interfered with testing locally. I wonder if ajv isn't resolving the relative id we have as a temp solution?

mattgarrish:
Although it looks like ajv does just treat them as identifiers according to this: https://github.com/ajv-validator/ajv#ref

But then I'm not sure why it would resolve the files in the same directory. Do you load them in your code? Does it need updating to /module there?

marisademeglio:
> Although it looks like ajv does just treat them as identifiers according to this: https://github.com/ajv-validator/ajv#ref
> 
> But then I'm not sure why it would resolve the files in the same directory. Do you load them in your code? Does it need updating to /module there?

Yeah, that's why - everything gets loaded before being handed to ajv. So it can only use ID and has no awareness of the filesystem. It's fine, I can just modify the schemas slightly before using them and then it works ok.

PR #227, opened 2020-07-06
Schemas fixes

mattgarrish:
A couple of additional things I've fixed in the schemas;

- We were using `uniqueItems: true` inconsistently and in many places where the specification says nothing about the need for uniqueness (even if it makes intuitive sense). You can declare a `type` more than once, for example, without negative effect. I'd suggest we leave messages about duplication for a full-fledged validator to handle, as these shouldn't be errors but more along the lines of epubcheck's info messages (best practices). The only required uniqueness is within `resources`, `readingOrder` and `links` and that check remains in the resource.categorization.schema.json file.
- I replaced two additional regexes for duration/language checking with a references to the patterns I broke out earlier.

### DIFF
--- a/schema/module/ItemList.schema.json
+++ b/schema/module/ItemList.schema.json
@@ -27,8 +27,7 @@
             ],
             "items": {
                 "type": "string"
-            },
-            "uniqueItems": true
+            }
         }
     },
     "required": [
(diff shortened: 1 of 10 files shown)


marisademeglio:
I get an error now that "type" should be an array, whereas before it could be an array or single string value.

PR #229, opened 2020-08-31
reading order is required in the spec

llemeurfr:
I spotted that the audiobook schema has readingOrder required, not the generic pub-manifest schema. 
But readingOrder is required by the pub-manifest spec.

### DIFF
--- a/schema/publication.schema.json
+++ b/schema/publication.schema.json
@@ -119,6 +119,7 @@
 	},
 	"required": [
 		"@context",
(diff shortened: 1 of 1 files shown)


mattgarrish:
A reading order isn't required in the authored manifest. If it's not present, the page that references the manifest becomes the sole entry after processing.

I expect readingOrder is required in the audiobooks schema because there always has to be at least one audio resource, so the same fallback can't/won't apply.

PR #233, opened 2020-09-10
Fix to use correct variable name

naglis:
From the context (`normalized` should be a map here), I think it should be `normalized`, not `item`.

***
<a href="https://pr-preview.s3.amazonaws.com/naglis/pub-manifest/pull/233.html" title="Last updated on Sep 10, 2020, 11:54 PM UTC (a8255c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/233/50cbb97...naglis:a8255c7.html" title="Last updated on Sep 10, 2020, 11:54 PM UTC (a8255c7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3487,7 +3487,7 @@
 											href="https://infra.spec.whatwg.org/#ordered-map">map</a>:</p>
 									<ol>
 										<li>
-											<p>if <var>item["type"]</var> is set and includes a <a>recognized type</a>,
+											<p>if <var>normalized["type"]</var> is set and includes a <a>recognized type</a>,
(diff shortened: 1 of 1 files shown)


iherman:

iherman (approved):
I believe there is indeed a bug there but, when it comes to a bug in the processing algorithm, I'd prefer the approval of the other editor, too. @mattgarrish?

mattgarrish:
Looks like a copy-paste error. `item` is only defined for iterating the list, so for the map case, yes, it should just be `normalized["type"]`.

PR #234, opened 2020-09-12
Fix variable in recursive normalization

naglis:
At 7.4.1 step 9.a.i, if `normalized` is a list containing `item` map, e.g.:

```json
[
    {
        "type": [
            "Person"
        ],
        "name": "Herman Melville"
    }
]
```

it seems that context for the recursive normalize data call should be `item["type"]`, not `keyValue["item"]`. In this case, context would be `Person`.

Same applies for step 9.b.i.

It appears `item["type"]` is being used in the reference implementation:

https://github.com/w3c/pub-manifest/blob/9a17073f65c5bae1e120933a9676250016b70148/experiments/manifest-to-internal-processor/manifestProcessor.js#L624

***
<a href="https://pr-preview.s3.amazonaws.com/naglis/pub-manifest/pull/234.html" title="Last updated on Sep 12, 2020, 10:41 PM UTC (3b6fe5a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/234/9a17073...naglis:3b6fe5a.html" title="Last updated on Sep 12, 2020, 10:41 PM UTC (3b6fe5a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3473,7 +3473,7 @@
 													href="#processing-normalize">normalize data</a>, when successful,
 												given <var>key</var>, <var>keyValue</var>, <var>lang</var>,
 													<var>dir</var>, <var>base</var> and using
-													<var>keyValue["type"]</var> as the context. If failure is returned,
+													<var>item["type"]</var> as the context. If failure is returned,
 													<a href="https://infra.spec.whatwg.org/#list-remove">remove</a>
 												<var>key</var> from <var>item</var>.</p>
 										</li>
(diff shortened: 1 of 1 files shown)


iherman:

iherman (commented):
It is difficult for me to 100% judge this based on my implementation (because it is in Typescript, the concept of "type" is more implicit based on the typing feature of TS), but I believe the proposed changes are indeed right.

mattgarrish:
Looks right to me.

In 9.a.1, `keyValue` is the value of the key, so unless it's a map with a type it doesn't make any sense. It's the type of map that is sent in with each key/value pair. Same with 9.b.1.

PR #242, opened 2020-10-01
Unify map key syntax

### LINKED ISSUE #260: Replace uses of dot notation
The user agent processing section refers to the variable `toc.entries` but everywhere else we use bracket notation for referring to variables. These instances should be changed to `toc["entries"]`.

naglis:
Here dot notation is used, however bracket notation is used elsewhere throughout the document, so I think it is worth to make it consistent.

Fixes #260 

***
<a href="https://pr-preview.s3.amazonaws.com/naglis/pub-manifest/pull/242.html" title="Last updated on Feb 4, 2021, 3:28 PM UTC (c04e298)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/242/f63fe7c...naglis:c04e298.html" title="Last updated on Feb 4, 2021, 3:28 PM UTC (c04e298)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4970,7 +4970,7 @@ dictionary LocalizableString {
 												<var>branches</var> and set <var>current_toc_node</var> to it.</p>
 									</li>
 									<li>
-										<p>Otherwise, if <var>toc.entries</var> contains an empty <a
+										<p>Otherwise, if <var>toc["entries"]</var> contains an empty <a
 												href="https://infra.spec.whatwg.org/#list">list</a>, set it to <a
 												href="https://infra.spec.whatwg.org/#nulls">null</a>.</p>
 									</li>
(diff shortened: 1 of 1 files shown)


naglis:

iherman:
@naglis, the documents are now in Proposed Recommendation. This means that, from now on, no changes should be made to the document; what we have is what will be published as a Recommendation. #243 is a borderline (spelling mistake) that can be done when we publish the Rec; but even this change is, I think, off limit.

From now on the mechanism is to label such issues or PR-s as an Erratum; when the maintenance group comes into life, that WG may decide to periodically republish the document with errata changes when these should be taken into account.

Cc @mattgarrish @wareid

mattgarrish:
I've added #260 so we have a reference for this pull request.

iherman (approved):

PR #245, opened 2020-11-04
Fix bidi example

### LINKED ISSUE #244: RTL text examples lose their RTL marker
The RTL text in examples, despite markup like
```
someText {
  value : <bdo dir="rtl">"HTML היא שפת סימון."</bdo>
```

isn't marked as RTL in the final rendered docs. My first guess is that respec (perhaps through the code prettifier) is removing the markup.

mattgarrish:
This fixes #244 by adding a parenthetical to the note to explain why the example text differs. It also adds the `bdo` tag to the REC snapshot to fix the rendering.

Note that the editor's draft will still strip the `bdo` tag, so to preview the fully-updated section, please see the snapshot: https://raw.githack.com/w3c/pub-manifest/editorial/text-dir/snapshot/index.html#example-2-set-the-language-and-the-base-direction-of-a-string

Let me know if the added para can use any improvement.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/pull/245.html" title="Last updated on Nov 5, 2020, 1:37 PM UTC (fe3102c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pub-manifest/245/b5c50f7...fe3102c.html" title="Last updated on Nov 5, 2020, 1:37 PM UTC (fe3102c)">Diff</a>

### DIFF
--- a/snapshot/index.html
+++ b/snapshot/index.html
@@ -1849,7 +1849,7 @@
 								<span class="example-title">: Set the language and the base direction of a string</span>
 							</div>
 							<pre aria-busy="false"><code class="hljs json">{
-    <span class="hljs-attr">"value"</span>     : <span class="hljs-string">"HTML היא שפת סימון."</span>,
+    <span class="hljs-attr">"value"</span>     : <span class="hljs-string">"<bdo dir="ltr">HTML היא שפת סימון.</bdo>"</span>,
     <span class="hljs-attr">"language"</span>  : <span class="hljs-string">"he"</span>,
     <span class="hljs-attr">"direction"</span> : <span class="hljs-string">"rtl"</span>
 }</code></pre>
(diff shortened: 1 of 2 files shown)


snapshot/index.html
```diff
@@ -1859,15 +1859,18 @@ <h5 id="x4-2-4-1-localizable-strings"><bdi class="secno">4.2.4.1 </bdi>Localizab
...
 									presence of a Latin character starting the string, as: </p>
 								<p dir="ltr">HTML היא שפת סימון.</p>
-								<p> However, that would be incorrect. The extra <code>direction</code> value is
-									necessary to control the display to yield: </p>
+								<p>However, that would be incorrect. The extra <code>direction</code> value is necessary
+									to control the display to yield: </p>
 								<p dir="rtl">HTML היא שפת סימון.</p>
+								<p>(Note that the <var>value</var> field in the example represents the text as it is
+									stored in memory, hence the discrepancy between it and the two renderings depicted
+									here.)</p>
```
iherman, line 1873: I may also add something like that in the note: "Text editors may also display the text using the Unicode Bidirectional Algorithm only, instead of the memory order."

iherman (approved):

snapshot/index.html
r12a: @iherman I'm not sure what you're trying to say here. When you say 'the text', which text are you referring to?  Do you mean that if you try to cut and past the text of the example you will lose the bdo ordering? Note btw that the first example in the note does exactly what you seem to be describing here.

mattgarrish: Here's how I've reworded the paragraph:

> Note that the value field in the example represents the text as it is stored in memory, hence the discrepancy between it and the two renderings depicted here. Text editors might also display the JSON value differently (e.g., using the Unicode Bidirectional Algorithm only).

iherman: If I cut-and-paste the right text (ie, the one with the HTML on the right hand side and properly display the Hebrew text), and then put it into a json file in my screen editor, what I get is the first example in the note and not the memory order as in our example...

PR #248, opened 2021-06-17
Moved to github actions

iherman:
The yml script in .github/workflows is now a dummy; if it is to be used then it should be a .yml file and the spec status should be updated.


.github/workflows/auto-publish.this_should_be_yml
```diff
@@ -0,0 +1,19 @@
+# .github/workflows/pr-push.yml
+name: CI
+on:
+  pull_request: {}
+  push:
+    branches: [main]
```
plehegar, line 6: previously, this was only done on the branch "publish_wd". Are you sure you want main? (it seems so but want to be sure)

.github/workflows/auto-publish.this_should_be_yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: w3c/spec-prod@v2
+        with:
+          W3C_ECHIDNA_TOKEN: ${{ secrets.W3C_TR_TOKEN }}
+          W3C_WG_DECISION_URL: https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-08-19-pwg#resolution3
+          W3C_BUILD_OVERRIDE: |
+             shortName: pub-manifest
+             specStatus: REC
```
plehegar, line 19: This is fine to leave as-is until we actually supports REC publications through Echidna.

plehegar (approved):
Also, did you actually remove the file ECHIDNA? The PR thinks you simply renamed it...

.github/workflows/auto-publish.this_should_be_yml
iherman: Yeah, I think doing it simple is fine. The repo is dormant anyway, so if we wake up, this can be changed if we want.

.github/workflows/auto-publish.this_should_be_yml
iherman: Yes, again, this is to be updated in case we reopen the wg...

iherman:
> Also, did you actually remove the file ECHIDNA? The PR thinks you simply renamed it...

The ECHIDNA file was part of a separate 'snapshot' folder which was used to ECHIDNA. Ie, the editor did all kinds of tests himself before publishing. So the file was actually moved, which means renaming...

plehegar:
ok

plehegar (approved):

PR #262, opened 2023-10-02
Update the respec structure to refer to the `main` branch instead of `master`

iherman:
This is the counterpart of https://github.com/w3c/audiobooks/pull/119

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,7 +47,7 @@
 				wgPublicList: "public-pm-wg",
 				github: {
 					repoURL: "https://github.com/w3c/pub-manifest/",
-					branch: "master"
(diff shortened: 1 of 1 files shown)


mattgarrish (approved):

iherman:
I suppose this (and the audiobook one) can be merged without further ado, there is no need to add a change log, right @mattgarrish ?

mattgarrish:
Right, my understanding is that we will only add change log entries for substantive changes from here on - class 3 and 4.

This is only a class 2 change, and arguably not even that since I wouldn't consider the spec info part of the text of the specification.

PR #263, opened 2023-10-05
Inline definition of sectioning root elements

mattgarrish:
This does the base minimum to address https://github.com/w3c/pm-wg/issues/3 -- I've inlined the list of elements as mentioned in https://github.com/w3c/pm-wg/issues/3#issuecomment-1723297359

The mix of elements is still odd (`body` can't occur inside of the body, etc.), and the spec still talks about outlines in relation to these elements, but I'll open a separate issue for those problems.

Fixes https://github.com/w3c/pm-wg/issues/3

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,7 +22,6 @@
 				previousMaturity: "PR",
 				implementationReportURI: "https://www.w3.org/publishing/groups/publ-wg/implementation/results.html",
 				crEnd: "2020-03-31",
-				updateableRec: true,
 				edDraftURI: "https://w3c.github.io/pub-manifest/",
 				errata: "https://w3c.github.io/pub-manifest/errata/",
 				editors:[ {
@@ -72,7 +71,7 @@
 				publications, serialized in [[json-ld11]], to enable interoperability between publishing formats while
 				accommodating variances in the information that needs to be expressed.</p>
 		</section>
-		<section id="sotd"></section>
+		<section id="sotd" class="updateable-rec"></section>
 		<section id="intro">
 			<h2>Introduction</h2>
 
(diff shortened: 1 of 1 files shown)


iherman (approved):
I guess this is a class 2 change, right? I have labeled it as such... (nothing really changes, just the reference change in some way)

mattgarrish:
Should we go ahead and merge this, since it's not changing any existing requirements, or do you think we should wait and get a group resolution on the next call?

iherman:
The issue is of class 3. I would think that a more explicit approval of the WG is necessary. @wareid @TzviyaSiegman @shiestyle, maybe we could make an email review call to approve changes of this sort. This seems to have worked before.

On the editorial side: @mattgarrish, I am not sure how you intend to manage this in general, but shouldn't the document be prepared for the candidate correction phase? That is what class 3 requires...

iherman:
I just see that the PR is marked as class 2, whereas #3 is set as class 3. Which one is correct? This should be in sync...

mattgarrish:
We're not fully addressing what is in the pm-wg tracker issue, so I don't know what needs to be done there. It maybe stays open as class three but refers now to #264.

This is the change we discussed in that thread that avoids us making a class 3 change until we have more input on how to change the toc processing algorithm (i.e., it just inlines the now-missing element list from html). Fixing #264 will fix the pm-wg issue.

iherman:
@mattgarrish yes, I remember now. The class labeling has been removed from  https://github.com/w3c/pm-wg/issues/3 now, and with that the issue in https://github.com/w3c/pub-manifest/pull/263#issuecomment-1803285090 is moot...

Answering to your original question with a fresh eye: in my view yes, it could be merged. But our fearless chairs should also give their opinion :-)

iherman:
@mattgarrish yes, I remember now.  https://github.com/w3c/pub-manifest/pull/263#issuecomment-1803285090 is moot... but then we should not make https://github.com/w3c/pm-wg/issues/3 automatically closed if this PR is merged.

Answering to your original question with a fresh eye: in my view yes, it could be merged. But our fearless chairs should also give their opinion :-)

mattgarrish:
> we should not make [w3c/pm-wg#3](https://github.com/w3c/pm-wg/issues/3) automatically closed if this PR is merged.

Right, I've stripped the reference, although I'm not sure it would have worked as I don't think I have access to close issues across to that repository.

iherman:
> > we should not make [w3c/pm-wg#3](https://github.com/w3c/pm-wg/issues/3) automatically closed if this PR is merged.
> 
> Right, I've stripped the reference, although I'm not sure it would have worked as I don't think I have access to close issues across to that repository.

Don't know either, but I have just given you "maintain" rights on that repo as well, just in case...

